PR #155, opened 2015-01-10
Plurals support

paradix:


jonasoreland:
Can you please change so that only lines you actually changed shows up in diff.
Currently 1848 shows up in diff, although you only modified a fraction of that.
The extra 1500 makes it much harder to see the changes you made.

paradix:
Ok, will do that, but that just indentations with 4 spaces (must have had the auto-indent set).

jonasoreland:
thx. now it's small enough that I can have "real" comments:
1) looks mainly good
2) for values-tr, are you sure it works to just put quantity "one",
it would feel safe to put the same value also on "many"
what do you think ?
3) cueDistance
you seem to have removed handling of km/miles
you shouldn't have to use the strings "km" or "mi"
you format code like "if(" which should be "if ("
in short this function looks broken.

paradix:
Ad 2) From what I read plurals in Turkish only require the quantity "one"
Ad 3) Will be fixed with next commit

jonasoreland:
1) one think you missed,
you should be using R.string.metrics_distance_km (or R.string.metrics_distance_mi)
instead of strings "km" and "mi"

2) I read the documentation of plurals, and I think "other" feels much safer to use than "one"

/Jonas

jonasoreland:
i made last changes myself...

paradix:
Ok. Just came home, but I'm too late :)

paradix:
I saw You pushed the plurals to Transifex, but at least for polish they seem to be not translated.
Did You expect this, or did just something go wrong?

jonasoreland:
something wrong, tried again...
does it look better ?

paradix:
polish translations for plurals are not present in Transifex. Other languages are fine.

jonasoreland:
can you fix (in trasferix) and i'll pull it down to git ?

/Jonas

paradix:
Ohh other languages are also missing the plurals translations. Something must have gone wrong :/

jonasoreland:
sure ?
but I modified the turkish myself...

do you mean that they are ok in git, but wrong in transferix ?

paradix:
Yes I'll fix for all languages I know in Transifex and let you know. If not today than early in the morning tomorrow.

paradix:
One sec I'll double check

paradix:
E.g. cues.xml for German in git has all the plurals translated. On the other hand Transifex for German (de) is missing all plurals, and for German (de_DE) is missing half of the plurals <-- weird ?

jonasoreland:
really weird...
do you think you can push all of git to transerix,
or should I try too ?

/Jonas

paradix:
I'll do it based on github. No problem for me.

jonasoreland:
oki, thanks!

/Jonas

paradix:
I think I have only right to edit Polish in Transifex.

jonasoreland:
so that might be the problem...
i don't have all the permissions either...

can we mail the other translators ?? (i have no idea who they are)

/Jonas

paradix:
Who did setup the transifex account fot You app? He/She should have all the persmissions

paradix:
This one (https://www.transifex.com/accounts/profile/Cqnqrd/) is mentioned on the Transifex page for Your app.

jonasoreland:
yes, i think it was him that started it...
i'll send a message to him...

/Jonas

jonasoreland:
pushed again now...
all of it...
can you check if it looks ok

/Jonas

paradix:
Ok, will check and give you feedback.

paradix:
German is now OK, but Polish is still missing the plurals

paradix:
Additionally Transifex sets different quantities e.g. in Polish "many" is replaced by "other".
So You should probably push from git to tx and not the other way around.

jonasoreland:
should you report that on
https://bitbucket.org/indifex/transifex/issues?status=new&status=open ??

/Jonas

PR #157, opened 2015-01-12
Update build.gradle to reflect antpluginlib update

shatteringlass:
This is bound to get broken again, though. Either we use it as a submodule or its link will break again and a new commit will be needed everytime.


jonasoreland:
i'll investigate git submodule when i find time...
/Jonas

jonasoreland:
i think i'll managed to make ANT+ into submodule
i'll close this.

Issue #153, opened 2015-01-06
Integrate gramaticly correct cues in polish


paradix:
Polish language has some linguistic specifics when it comes to countable nouns.
We differentiate between nouns for 1, 2-4 and 5 up.

I would like to implement this behaviour for cues in the Formatter class.

jonasoreland:
Good!
See this:
http://developer.android.com/guide/topics/resources/string-resource.html#Plurals

jonasoreland:
please give example

paradix:
Ok, so right now the cues are constructed like this:

activity 8 meters
activity 20 minutes 42 seconds
activity 3 minutes 11 seconds per kilometer
activity 166 bpm
activity heart rate zone 3.4
interval 306 meters
interval 2 minutes 11 seconds
interval 1 minute 12 seconds per kilometer
interval 173 bpm
interval heart rate zone 3.5
lap 690 meters
lap 2 minutes 24 seconds
lap 10 minutes 27 seconds per kilometer
lap 171 bpm
lap heart rate zone 3.4

E.g. in polish it's almost impossible :) to create a sensible-sounding cue with just one word preceding the actual result.
So I was thinking to add the dimension into the cue (like in the HRZ cue), and then the cues would be like this:
activity distance 8 meters
activity time 20 minutes 42 seconds
activity pace 3 minutes 11 seconds per kilometer
activity heart rate 166 bpm
activity heart rate zone 3.4
...

jonasoreland:
that sounds like a good plan.

but i think it should be done in strings.xml somehow,
cause for me (in english) the dimension is just redundant extra noise.

at least I want to cues to be short and distinct, even if they aren't
necessarily grammatically correct.

what do you think ?

/Jonas

paradix:
Well I absolutely get You point of view, but it won't be possible through strings.xml.
Now each cue consists of the Scope.cueId + the value of the dimension. The Scope.cueId is one string value for each enum and in You proposal I would need multiple string values for one cueId depending on which dimension is being processed.

What would You thing about this proposals.

Activity level:
Distance 8 meters
Time 20 minutes 42 seconds
Pace 3 minutes 11 seconds per kilometer
Heart rate 166 bpm
Heart rate zone 3.4

Interval level:
Interval distance 306 meters
Interval time 2 minutes 11 seconds
Interval pace 1 minute 12 seconds per kilometer
Interval heart rate 173 bpm
Interval heart rate zone 3.5

Lap level:
[...]

PR #170, opened 2015-01-16
final proposal for Issue #153

paradix:
Could You please make a review and tell me if this solution is acceptable?
If yes, I will than finish it for all other languages and make some clean up in the code.

### DIFF
--- a/app/res/values-pl/cues.xml
+++ b/app/res/values-pl/cues.xml
@@ -14,20 +14,23 @@
   ~
   ~  You should have received a copy of the GNU General Public License
   ~  along with this program.  If not, see <http://www.gnu.org/licenses/>.-->
-  <string name="cue_workout">aktywność</string>
-  <string name="cue_activity">interwał</string>
-  <string name="cue_lap">okrążenie</string>
+  <string name="cue_interval_pattern">{1} {0} {2}</string>
+  <string name="cue_event_pattern">{0} {1}</string>
+  <string name="cue_workout">10#Treningu|20#Trening</string>
+  <string name="cue_activity">10#Interwału|20#interwał</string>
+  <string name="cue_lap">10#Okrążenia|20#Okrążenie</string>
+  <string name="cue_dimension_pattern">1#{3}|2#{3}|3#{3}|4#{3}|5#{3}|6#{3}</string>
   <string name="cue_warmup">rozgrzewka</string>
   <string name="cue_cooldown">uspokojenie</string>
   <string name="cue_current">aktualny</string>
   <string name="cue_recovery">regeneracja</string>
   <string name="cue_perkilometer">na kilometr</string>
   <string name="cue_permile">na milę</string>
-  <string name="cue_started">rozpoczęte</string>
-  <string name="cue_paused">zatrzymana</string>
-  <string name="cue_resumed">wznowiona</string>
-  <string name="cue_stopped">zatrzymana</string>
-  <string name="cue_completed">zakończony</string>
+  <string name="cue_started">101#rozpoczęty|102#rozpoczęty|103#rozpoczęte|106#rozpoczęta|109#rozpoczęte</string>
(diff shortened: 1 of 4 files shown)


paradix:
There are still some comments missing that will be needed for translators. But I will add them after You commented on this.

jonasoreland:
Hi,

I'm not happy with the PR, sorry (but still like the concept!)

1) the existing cues should be unaltered (as strings, not MessageFormats)
2) the *sentences* should get new entries in cues.xml, they can be MessageFormat
3) new formatting preferably be in Formatter not in AudioFeedback

Don't you think above would work ?

/Jonas

paradix:
Hi Jonas,

I get ghe point with moving the logic to the Fotmatter class, but I don't understand why You want the existing cues stay unaltered.

What will be the use of them? In this form they are useless e.g. for the Polish translation, as it needs many forms of verbs an nouns even for such easy sentences.

The whole point of my idea was to make an flexible mechanism that would leave the translator some space in creating the cue sentence.

The result is unchanged for English and Polish can be handled. I assume other languages can make use of this as well.

Regards
paradix

jonasoreland:
My motivation for wanting the existing cues unaltered is
- it is significantly easier to translate single words
- even if this makes certain (all) languages grammatically incorrect, the
effort of getting a reasonably good baseline is decreased.
- only few experts will use actually translate whole sentances

do you see what I mean ?

/Jonas

paradix:
with that proposal You can leave the existing cues in this format:
<string name="cue_workout">0&#60;activity</string>
<string name="cue_activity">0&#60;interval</string>
<string name="cue_lap">0&#60;lap</string>
<string name="cue_started">0&#60;started</string>
<string name="cue_paused">0&#60;paused</string>
<string name="cue_resumed">0&#60;resumed</string>
<string name="cue_stopped">0&#60;stopped</string>
<string name="cue_completed">0&#60;completed</string>

It still would work, and the strings would be altered minimalistic. 
Would that be a way out for You?

jonasoreland:
not really,

what do you see as the downside of adding new strings for the sentences ? I
don't understand
(i think it's quite logical, strings for nouns/verbs and strings for
sentences)

/Jonas

paradix:
The problem is, that in other languages then English due to declension you might need more then one form of the noun, and due to conjugation more than one form of the verb.

Especially in Polish you have 7 different forms in the declension.
So with the format I have proposed you don't need to care that much about this, because the translator will have to make the proper decision.

You can't know how many strings you will have to add for nouns and verbs, as it is language dependant. 

That's my point of view, but maybe I don't understand Your proposal.

jonasoreland:
1) i'm more happy if i get 10 more languages translated, with grammatical
errors,
that if i get polish to handle 7 different forms of declension and
conjugation.

2) as I see if, if you make the it easy to do an "easy" translation, you
make the complicated infinitively complicated.
i don't mind if you enumerate each combination of dimension/scope for
different values.

so...handle more of polish FINE! but not that the expense of complicating
it for those that doesn't want to...

---

does that make sense ??

/Jonas

paradix:
Sorry I didn't mean to start a fight over this. Just thought this might actually be some added value to get it right. And with the option in my last comment it is not getting any more complicated for other languages. You would need to insert the substring "0<" into each cue.

But in the end it's up to you, as it is your application. Leaving the polish cues as they are translated now will still be understandable though not really grammatically correct.

jonasoreland:
no worries. different opinions/perspective is good.
i'll look at your latest version tomorrow....

/Jonas

paradix:
OK, I will at least push a commit tomorrow with a re-factored version so you can see it all

paradix:
Ok, now have a look. I have re-factored it, so most of the logic is now in the formatter.
Plus the cues.xml have been altered for all languages so you can see the whole scope of the changes.

gerhardol:
Is this PR obsolete with the changes to nouns in 1.57?

gerhardol:
Plurals are handled in v1.58
Is that enough to close this PR?

PR #172, opened 2015-01-19
zone distribution bar

paradix:


jonasoreland:
Hi,

I tried this on my samsung galaxy s4 mini.
1) Why the white/grey background ?
2) Could it be made smaller ? (it almost doesn't fit my screen)
maybe you can try in on small size screen in an emulator ??
3) When having a very large percentage in one zone,
the red bar over writes the percentage number.

But, it looks quite fancy!

/Jonas

paradix:
1) I will try other background colours maybe black ;)
2) it should fit all the space between notes and the buttons. But maybe setting a maxHeight wouldn't be so bad. 
3) that's weird, because it shouldn't based on what I was thinking. I'll check that out.

paradix:
1) changed
2) weird I can't figure out the right height of the hrzonesBarLayout so the graph hides underneath the buttons. Any idea?
3) fixed

PR #178, opened 2015-01-25
Download Nike+ Activities

vyaslav:
This is not finished but maybe someone can help with fixing Graph view for downloaded activity


jonasoreland:
preliminary comments:
1) rename directory "/activity" to "/entity" or something like that.
2) move nike code into nike uploader, all code that is nike specific should be there.
3) delete ExternalActivitySerializer
4) rename SportActivity to Activity (matching table)
5) remove changes from DBHhelper, 
the new columns aren't needed, they already exist.
6) put downloader functionality inside Uploader (but you might think of a new name for the two)
7) all over, follow coding standard e.g "if(" => "if ("...

puhh....there is more, but this is all for now

vyaslav:
Thanks, most of the comments make totally sense for me.
But I have few question:
3) It is used by AccountActivity to abstract from specific serializers and handle generic logic why should it be removed? 
5) I haven't found in activity fields where the source of activity like Nike and its external id is stored, maybe only for feed which is not suitable. Can you please point me out?
6) Could you please explain why should we avoid the power of interfaces? From my point of view it is logical to have methods for downloading and uploading separated instead of putting everything related and not related to one class.

Hope you will help me understand issues with graph as it was the intention of this PR.

jonasoreland:
3) i don't think class adds anything valueable
I don't see why json should be written to file in the first place
And if it should, i don't see that it has too for all uploaders
5) DB.EXPORT.TABLE
it's a relation table since one activity can be uploaded to several accounts.
6) i think they (uploader/downloader) are related,
typically they share knowledge about how the report site works, it's data-format, it's login handling

paradix:
I would call the class that handles different fitness services something like "...Synchronizer". 
This class should handle all functionality connected with the specified service.

vyaslav:
3) It wasn't for uploader only for downloader which is optional, the reason to keep json file for me personaly is to when new features are available I can just reparse existing files instead of trying downloading the again and also External API can change... Also I think having GPS data in sqllite database is not so efficient because its accessed not so often but can slow down a queries because of rows count... Anyway I will remove it if it such a paint point for you.
5) Now I found it, thanks. But I'm not getting why it is called CREATE_TABLE_REPORT, I thought I should use it for some reports...
6) I find it is nice idea to rename Uploader class to Synchronizer but I would keep two interfaces as contract here is deferent, I'm agree that implementation shares some connectivity methods, but its not an implementation, its interfaces

Issue #173, opened 2015-01-20
Feature - Advanced workout creator/editor


paradix:
Any interest in a possibility to create advanced workouts off-line within the app? 
Right now they can be downloaded from garmin connect but there's no way to create them inside the app, but not everyone has/wants to have an garmin account.

Issue #158, opened 2015-01-12
Integration with google fit?


j-xella:

jonasoreland:
patches accepted!

paradix:
I'll take a look at this one

paradix:
Do you prefer using the REST API or the Fitness API? 
But I think the Fitness API requires that the app is registered and have the API enabled.

jonasoreland:
1) i prefer REST (since all of the other integration is made that way).

2) i think i need to register the app anyway...do you know where the page
is ?

there will most likely be a app id/app secret that one get...
and these i have encrypted/obfuscated for other uploaders
and give them in the constructor...

/Jonas

paradix:
I think you will find all info here https://developers.google.com/fit/android/get-started

jonasoreland:
hmm...enabled fitness api.
what do I need to do more ?


/Jonas

paradix:
Give me some time, I need to figure this out :) and will let you know if it works.

jonasoreland:
what is your private email, and I can mail you oauth2 info...(suitable for
REST)

/Jonas

KaptenJansson:
As a workaround, use a tracker like Runkeeper as a proxy (it has Google fit integration).

paradix:
Hi Jonas,
I already did the part with the refresh token, so authorization works fine.
But I have a problem with creating data sources that are needed for google fit to work.
Somehow the POST request gives me a 400 response and I can't get it to work.
Maybe you could have a look at my branch at commit a01cdead57a5f94dba5719b49338d03bab0d103c

paradix:
seems my json is malformed, don't waste your time for now :)

paradix:
I'm pretty much done with the coding, except one thing.
The HttpURLConnection doesn't  support the PATCH request.
Do you have any idea how this can be done?

jonasoreland:
have you tried

X-HTTP-Method-Override: PATCH

http://stackoverflow.com/questions/19797842/patch-request-android-volley

paradix:
Great this worked, but still getting a bad request on the POST and the response doesn't help much.

paradix:
Could you please email me the "developer project number" just to check if I have the proper one. Because it seams that the number is not being accepted by google api when I try to post the activity.

paradix:
I'm stuck :/
I think that code-wise the google fit uploader to 90% ready/working and can be tested, but.
[...]

PR #179, opened 2015-01-26
Advanced workout creator

paradix:
New Feature - advanced workout creator (fixed issue #173).
 - Includes new activity for creating advanced workouts that can be started from workout manager.
 - You can add steps and repeat steps by clicking the proper button.
 - Adding steps underneath the repeat step can be done by a long-click
 - Deleting of steps is done by a long-click

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -14,173 +14,162 @@
   ~
   ~  You should have received a copy of the GNU General Public License
   ~  along with this program.  If not, see <http://www.gnu.org/licenses/>.
--->
+  -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    package="org.runnerup" >
+	package="org.runnerup"
+	xmlns:tools="http://schemas.android.com/tools">
 
-    <uses-sdk android:minSdkVersion="8" />
+        <uses-sdk
+            android:minSdkVersion="8" />
 
-    <uses-feature
-        android:glEsVersion="0x00020000"
-        android:required="true" />
+		<uses-feature
+  			android:glEsVersion="0x00020000"
+  			android:required="true"/>
 
-    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
-    <uses-permission android:name="android.permission.ACCESS_LOCATION_EXTRA_COMMANDS" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
-    <uses-permission android:name="com.google.android.providers.gsf.permission.READ_GSERVICES" />
-    <uses-permission android:name="android.permission.WAKE_LOCK" />
-    <uses-permission android:name="android.permission.INTERNET" />
+		<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+		<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+		<uses-permission android:name="android.permission.ACCESS_LOCATION_EXTRA_COMMANDS" />
+		<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
+		<uses-permission android:name="com.google.android.providers.gsf.permission.READ_GSERVICES"/>
+		<uses-permission android:name="android.permission.WAKE_LOCK" />
+		<uses-permission android:name="android.permission.INTERNET" />
+		<permission android:name="org.runnerup.permission.MAPS_RECEIVE" android:protectionLevel="signature"/>
+        <uses-permission android:name="org.runnerup.permission.MAPS_RECEIVE"/>
+    	<uses-permission android:name="android.permission.BLUETOOTH" />
+    	<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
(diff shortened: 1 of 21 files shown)


jonasoreland:
fyi: glanced, and all look good.
i'll pull and test it some manually before merging though.

/Jonas

paradix:
Great, I have already made some manual tests. Check please if the changes made to the WorkoutSerializer.createJSON don't mess with the workouts downloaded from garmin, because I was only able to test with the one supplied with the app.

jonasoreland:
comments (after test)
1) i can scroll down the list *under* the buttons on the manage workout screen
2) i get yes/no instead of OK/cancel when I create a workout (set name)
3) how do I add a step under a repeat (i.e so that it's repeated)
4) it's (to me) totally unobvious that long press means delete of step
and...no "are you sure" when it happens.
5) I expected save to close the edit-screen, shouldn't it ?

6) code wise: when are the changes to WorkoutSerializer.createJSON needed ?

/Jonas

paradix:
1), 2), 5) I'll check this issues and fix them.
3) Long click on the repeat and You get a popup.
4) yes it's not so obvious I also see it, but I couldn't find any other way to delete single steps. I'm not that handy in UI programming, maybe You would have an better solution?

jonasoreland:
GUI suggestions...

1) would it be possible to have a "+" button on the repeat-step to add a step to the "repeat"
and maybe have the same "+" under all existing steps to add a new "in the end".

2) along the same lines, a "-" button/icon e.g on the left of each step.button could be delete...
(with a "Do you want to delete this step: "OK", "OK - don't ask again", "Cancel" popup)
 
3) and...moving of steps would also be nice...

---

what do you think about those suggestions? good/bad hard/easy

paradix:
Well everything is doable - it's always the question of time and material :)
I will see what I can do about your suggestions - seems like a nice idea.

paradix:
Could you please tell me more about the "in the end" step? I didn't realize there can be something more added after each step.

jonasoreland:
i simply meant a "+" button/icon...pressing it would add a step

paradix:
Ahh ok, I just read your comment again and I finally understood what you meant :)

paradix:
now most of your suggestions are done. I left the moving part for later.
Waiting for further comment's

jonasoreland:
cool! that was fast!
will try to find time tonight to check it out...

/Jonas

jonasoreland:
Hi there.
Much better!

Comments:
1) can you make the indentation bigger so that it's more obvious which steps are repeated ?
(i think that is quite important)

2) not so important: would it make sense to remove the "+" from the right of a non-repeat step,
and instead put it underneath the last step in the sequence ?

3) not super important: the "+" and "-" buttons are dead ugly if you ask me.
How about a round background image (can be done with shape) in e.g green for add and red for remove. (think the new "material design" thing, but with runnerup standard color)

paradix:
1) Maybe sub-steps could get another shade of Gray (LOL) instead?
2) that was also my primary idea, but I thought this way you can easily add steps inbeetween
3) 100% right - will think about something

jonasoreland:
1) color change is fine by me too...
just something so that you see....now i missed it at first...and second time...and third time,
and then I saw it.

2) ok...fair enough

3) ok

/Jonas

paradix:
Can you give me any hints on how to create image's that can be used as buttons? Did You do it in Gimp or something similar?

jonasoreland:
how about the one in wear/src/main/res/drawable/circle.xml ?

basic shapes can be defined in xml, no need for actual bitmap.

/Jonas

paradix:
I didn't mean the circle button, but the colour of the step button

jonasoreland:
ahh...that is likely one of the title_spinner*.png i think...
but modifying these seems...not optimal i think.

maybe larger indentation is better then ??

/Jonas

paradix:
Yeah will probably do that, but this will lead to text wrapping and will make the StepButton twice that high.

jonasoreland:
commonly or only with double repeat ??
(or with which strings ??)

i use a background color in the "Advanced"-tab.
maybe that makes it easier to see repeat indentation...maybe :)

/Jonas

paradix:
It will probably be only a problem for other languages e.g. Polish :) The translated words for the intensity are being wrapped, that's why I was thinking about another colour of the StepButton/TitleSpinner for the repeated steps. 

This would require to create another png image in a different colour. That's why switching to holo theme would be nice, or is that out of the question?

jonasoreland:
i'd say that switching themes as part of this feature is out of the
question...
though, one(you?) can work on that as another PR

/Jonas

paradix:
I didn't mean this as a part of this issue, just in general. 
I'll finish this one probably with making the padding of sub-steps bigger.
After that I'll do some reading about the holo themes and see what can be done, but this will probably mean a lot of work.

jonasoreland:
right, sounds good.
i don't have anything "against" holo-theme...i just haven't spent time on
it...

https://github.com/jonasoreland/runnerup/issues/158 is one thing that might
be good for broader adoption...
if you're looking for work :)

/Jonas

btw, are you member of
https://plus.google.com/u/0/communities/112040124647770742805 ?
i think i'll do a new beta this weekend or so...

paradix:
Have you had any chance to review/test the last changes?

jonasoreland:
looked good
merged
and please tell when a new reviewable version is available...it's easy to miss.

Thanks! Great work

/Jonas

PR #195, opened 2015-02-13
Code review for Google Fit

paradix:
The uploader Is working. Sometimes google throws an 500 internal error response, so the upload fails.
Maybe it could be handled somehow, but I will do it after your do the first review.
Code also needs cleaning up and maybe some comments ;)


paradix:
Ok I think that's it. You can make the full review now.

jonasoreland:
looks very good!

questions/comments:
1) why increase version from 8 to 9, there is still ~4% users on 8 (froyo)
2) you don't need to increase db version to add a new uploader.
(and increasing db version make downgrade impossible)

---

BUT, I will probably not merge this before next release... there are so many thing in the pipe now.

/Jonas

paradix:
1) was by mistake, I didn't notice I have checked it in. This was when I was trying to use the google fitness api client. This can be changed back to what it was.
2) couldn't figure out how this can be done. If you show me I'll change it.

jonasoreland:
So I merged it...

A question: why doesn't it show distance (on the web).
It only shows active time...

note: this is the first time I use google fit...

/Jonas

paradix:
untitled

You can see the distance in the detail view of every activity.

note: I also never used it before ;)

PR #197, opened 2015-02-20
Entity proof of concept

paradix:
Hi Jonas,

here is a quick PoC. It's not compilable but you can see the idea.

Waiting for you thoughts.

Regards


jonasoreland:
i really not convinced 6,800 additions and 614 deletions and no new features.
why do you want this ??
/Jonas

paradix:
Well this changes a lot in the way you handle all the db calls. 
On one hand you get all the business objects represented as classes which make the code readable and understandable. On the other hand you have all the data access and manipulation wrapped and encapsulated in one place.

Through the classes included in this commit you get an easy way to create sql statements with speaking names for all the commands and all the table fields.

I understand you point of view, that you don't get new functionality with this change. Functionality in means of application features, but I think you get a safer way of data handling and manipulation.

Using this will also mean also a lot of refactoring in the current code.

nebmo:
And I guess using content provider would let us expose (in a positive way)
runnerup information to other applications? Or have I misunderstood the
concept of content providers?

jonasoreland:
i suggest starting with what was in the other guys patch.
significantly less code. same functionality...

a subsequent change _might_ be what is proposed here...time will tell.

sounds ok ?

/Jonas

paradix:
@nebmo : yes, this could also be used to achieve this. Wasn't the main idea but based on the concept of ContentProvider this could be done.

@jonasoreland : what is your main problem with the concept? More code doesn't make it bad. It just makes it well structured and follows the paradigms of oop ;)

jonasoreland:
more code is bad...

(repeat)

i suggest starting with what was in the other guys patch.
- same functionality (encapsulating entities in wrapper objects, "oop")
- significantly less code.

i also suggest adding entities for new code you write, and not alter the
world.

sounds ok ?

paradix:
Well I can't fight with the argument about lines of code. The other code isn't commented so obviously it has less lines. Just thought this could be an easy win in means of modularisation, reusability etc.
I see if I can do something with the other code

PR #202, opened 2015-02-26
Moved hr device code into its own module

xbao:
Hey, I'm wanting to use the HR device code in another application I'm writing. It's very self-contained, so moving it into it's own module to be included as an AAR was straight forward. The nice thing about the move for RunnerUp is all the build logic for the hr device stuff is out of the app's module. What do you think?

PR #205, opened 2015-02-28
Reverted pull #202

xbao:
Reverted pull #202 due to errors assembling release build. Issue is at xbao/runnerup#1 

Is this easier/better than you reverting it on your end? I'm relatively new to git and contributing, advice is always appreciated


xbao:
Shouldn't need this anymore, just need to add

```
-dontwarn com.samsung.**
```

to the proguard rules

PR #214, opened 2015-03-10
Basic entity concept

paradix:
Awaiting acceptance ;)


paradix:
Haha that was quick and easy

jonasoreland:
there a limit to how much time i can spend on it too...
i wasn't super keen on the implemented but it as KISS...and manageable in
size...

/Jonas

paradix:
So maybe find a bit more time later and let's do it in a way you want it to be done?

jonasoreland:
yea...or not :)

PR #216, opened 2015-03-11
Added HRData class

xbao:
Hello Jonas.

I've added an `HRData` class that represents a data point gathered from a device.

It provides a way to include information about R-R Intervals, if available. 

To access it I added the method `getHRData()` to `HRProvider`. 

This is intended to replace `getHRValue()` and `getHRValueTimestamp()`. This is good, as the values could be updated between the calls to those methods leading to inconsistent data or a race condition (I'm not sure what Java guarantees, but in C++ that can lead to torn reads if the value if half updated when the read occurs)

Let me know if you have any thoughts on this

Xiao


jonasoreland:
looks fine.
minor comments
1) "if (" notice space between "if" and "("
2) i really hate "if () return X" but much prefer if return is on own line

i think both of these happens if you format with AndroidStudio tool

paradix:
IMO you should use proper accessors methods for the fields and make the member variables private instead.

jonasoreland:
i actually disagree on the accessors,

that is class is only for passing around multiple data-points as one.
adding the accessors is IMO only useless bloat.

and if...one in the future for what ever unknown reason need to introduce
encapsulation, it's fine to do that then

/Jonas

xbao:
@jonasoreland 

1. No worries, I'll fix that up
2. I agree. I was following the Android style guide per the Contributing guidelines. It explicitly says braceless if's need to be on one line. I'll change that and let you know when it's done

@paradix I would agree if this class was anything other than a simple data object. With all these fancy IDE's It's super easy to change that if required in the future.

xbao:
@jonasoreland Done

Issue #215, opened 2015-03-11
Time and distance triggered cues doesn't work


paradix:
In version 1.43 on distance and time triggered cues doesn't work. 
No exceptions are being thrown just the TTS isn't speaking.

Can anyone confirm this?

nebmo:
I have only tried new lap cue and that seems to work
Den 11 mar 2015 18:21 skrev "paradix" <notifications@github.com>:

paradix:
Lap cue == "Lap started" etc?

nebmo:
I was talking about end of lap audio cue

jonasoreland:
i'll try in emulator tonight...

- what audio cue language do you use ?
- does the "test audio cue" button work ?

/Jonas

On Wed, Mar 11, 2015 at 10:49 PM, Niklas Weidemann <notifications@github.com

paradix:
I have used the polish language (I'll try others)
The button works, that's why I was wondering.

paradix:
Seems like my /data/data/org.runnerup/shared_prefs/Default_audio_cues.xml has disappeared.

paradix:
It looks like the problem present when the R.string.default_audio_scheme_name is translated
In this case the app looks for Default_audio_cues.xml which isn't present and expected would be to return the org.runnerup_preferences.xml PrefFile

paradix:
This can be closed now, I think.

PR #217, opened 2015-03-12
default audio scheme name

paradix:
Fixes issue #215

### DIFF
--- a/app/src/org/runnerup/view/AudioSchemeListAdapter.java
+++ b/app/src/org/runnerup/view/AudioSchemeListAdapter.java
@@ -57,8 +57,9 @@ class AudioSchemeListAdapter extends BaseAdapter {
 
     @Override
     public Object getItem(int position) {
-        if (position == 0)
-            return "Default";
+        if (position == 0) {
+            return inflater.getContext().getString(R.string.default_audio_scheme_name);
+        }
 
         position -= 1;
 


jonasoreland:
I like the patch...but does it really fix the problem ?
I.e if you change language on phone, your default audio cue scheme shouldn't change, right ?
Don't we need to handle that specially ?

paradix:
I'm still testing different issues with the language settings. We will see if I find any other problems.
For now the patch seem to work fine. I'll play around with the example you mention.

paradix:
This patch handles the default scheme correctly even if you change the phone language.

jonasoreland:
that is great! but how ?

paradix:
the default values of the audio cue prefs are not stored in a seperate file but in the runnerup global prefs. So if the name of the "default" scheme is equal to "default" the global prefs file is returned.

This comparison is made on the values of:
1) the preference key pref_basic_audio (the value of the scheme selected in the spinner)
AND
2) the value of the string default_audio_scheme_name

If both are accessed through the resources and get translated to the global locale all is fine.

The problem was in the spinner value returning always "Default" instead of the translated string. When the string default_audio_scheme_name was translated (as for polish) the comparison was returning false and the app assumed it's not the default but a custom audio scheme.

That's my explanation for this.

jonasoreland:
super, then I understand, thx!

/Jonas

PR #219, opened 2015-03-12
audio feedback language fixes #218

### LINKED ISSUE #218: AudioFeeback doesn't support audio language setting
The LocaleResources class works only for the Formatter class.
The AudioFeedback class still uses the global resources which gives false results ( translated in the global locale)

paradix:


jonasoreland:
but, isn't that dead code currently ??

paradix:
I think the two middle "else if" statements are never true. Also two constructors are never used.

paradix:
damn my check-in was wrong/

Issue #196, opened 2015-02-17
Downloading recorded activities from RunKeeper


paradix:
The main idea is to get a possibility to download previously recorded activities from Runkeeper into RunnerUp local database. 

With this solution there should be introduced a way of mapping database objects as entity object in form of Java classes (e.g. Lap, Activity, Location, etc.).

This way there will be a good code base created to make it possible to download activities from other serviced. 

Further argument for this solution
Based on my observation and my personal preferences I thought that it might be useful to create a tool that would keep different accounts in sync. Probably all (or at least many) users have accounts on different services at once just because we like different features of those services or because we just like to share our activities with many friends :)

With the proper design of the upper requirement, the code can be used to create another complementary app with synchronization functionality.

paradix:
Jonas what do you think about creating a RunnerUp database content provider based on the android.content.ContentProvider class. There are nice tools that make the process of generating appropriate java classes really easy?

jonasoreland:
don't know enough about it to have an opinion.

what is is used for ?
what are the tools ?
what should the generated classes be used for ?

generally, i'm not too keen on generated code either...

/Jonas

paradix:
Here are the basics: http://developer.android.com/reference/android/content/ContentProvider.html

The code is generated to make you life easier. You can/have to alter the code that will be generated. Based on a json representation the basic entities and all needed accessors will be automatically  available.

There will be no need for writing sql queries, all this will be wrapped in the ContentProvider.

jonasoreland:
preliminary opinion: not excited.
where do you want to use this ?
what is "wrong" with current proposal? i.e entities that are easy
construct, and we can write one class the persists them to db.

but...code is what matters, if you show me good code, i really don't care
how you wrote/generated it...

/Jonas

paradix:
In the end the ContenProvider is exactly what we were discussing earlier it just wraps it all up.
It's a pattern that can be used to manage the layer between our entities and the db.

Here you can find an example --> https://github.com/BoD/android-contentprovider-generator/tree/master/etc/sample/app/src/org/jraf/androidcontentprovidergenerator/sample/provider

In my opinion it's exactly what you want to achieve and you can get rid of all current queries to db that are in part plain text.

jonasoreland:
so...if you do pull-requests with good code that replaces current code, i'm
super happy!!

/Jonas

jonasoreland:
and if there are things that i don't know much about...
do a small proof a concept

/Jonas

[...]

PR #222, opened 2015-03-19
Download activities

paradix:
This are only changes around the RunKeeper downloader from Issue #196.
With this a base for all other services should be made.

### DIFF
--- a/app/res/layout/account.xml
+++ b/app/res/layout/account.xml
@@ -76,10 +76,20 @@
         android:background="@drawable/btn_blue"
         android:layout_width="fill_parent"
         android:layout_height="wrap_content"
-        android:layout_above="@+id/ok_account_button"
+        android:layout_above="@+id/account_download_button"
         android:text="@string/upload" />
-    
-    <Button
+
+  <Button
+    style="@style/ButtonText"
+    android:id="@+id/account_download_button"
+    android:background="@drawable/btn_blue"
+    android:layout_width="fill_parent"
+    android:layout_height="wrap_content"
+    android:layout_alignParentBottom="false"
+    android:text="Download"
+    android:layout_above="@+id/ok_account_button"/>
+
+  <Button
         android:id="@+id/ok_account_button"
         style="@style/ButtonText"
         android:background="@drawable/btn_green"
@@ -87,5 +97,5 @@
         android:layout_height="wrap_content"
         android:layout_alignParentBottom="true"
         android:text="@string/ok" />
-    
+
 </RelativeLayout>
--- a/app/res/layout/account.xml
+++ b/app/res/layout/account.xml
@@ -86,7 +86,7 @@
     android:layout_width="fill_parent"
     android:layout_height="wrap_content"
     android:layout_alignParentBottom="false"
-    android:text="Download"
+    android:text="@string/download"
     android:layout_above="@+id/ok_account_button"/>
 
   <Button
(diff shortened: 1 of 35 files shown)


jonasoreland:
i glanced it yesterday (all in 2 minutes)...and think it looks good.
i'll try find time to look more carefully during weekend

paradix:
OK, but don't merge it yet. I'm still testing and need to make sure no bugs were built in for the upload functionality.

paradix:
I think it should be now ready for review.

paradix:
I'm still tweaking on the activity compare method that is supposed to filter already present/downloaded activities.

jonasoreland:
i haven't had time to read your patch yet, sorry :(

BUT, the DB.EXPORT.EXTERNAL_ID in DB.EXPORT.TABLE is supposed to be the
external id of the activity for a specific uploader.
E.g in your case the runkeeper id for an activity.

If we modify the uploaders to populate that field...the we don't need to
compare activities.
And for simplicify, we can say that we only support downloading activities
that are newer than the version that support downloading.

Makes sense ?

/Jonas

paradix:
Well yes the use of the external_id seems like a good idea. But I'm assuming that not every activity has been uploaded with RunnerUp to every possible service. Hence there will not always be an external_id for comparison.

That's why I think it would be good to create some rules of similarity. 
For instance I have uploaded some activity created by RunnerUp to RunKeeper and I don't know why but RunKeeper says it's longer by few meter, or started a second later or the duration got messed up and is now of by some minutes.

We will probably newer get it 100% right, but close enough would be good enough.

jonasoreland:
general comments: i think splitting this PR into several smaller is a good idea.

first pass comments:
1) license is missing on "all" new files, please make a separate PR to fix that

2) What is IObjectValues, I couldn't figure out if I had 10 guesses...how about a name that describes it
class DBEntity
{
  interface Insertable {
  };

  interface Updatable {
 };
};

3) How about renaming Uploader to Syncronizer (in separate commit/PR)
4) How about creating a DefaultSynchronizer which implements the interface,
and let classes extend it (possibly via FormCrawler) so we don't add empty methods
all over when adding new capabilities.
5) I strongly think that "persistActivity" should not reside inside RunKeeperSynchronizer,
isn't that the entire idea with the DB entities, RunKeeperSyncronizer creates an object,
that will be inserted somewhere else.

6) random newlines/space/bracket changes here and there should be move to separate PR or removed all together

paradix:
I agree on all your point. I'll do the changes in the next iteration as there will be further refactoring necessary.

Ad 2) The idea was to create a common interface for the Db entities to be able to make bulk updates/inserts. The name obviously could be more speaking.

Ad 5) Yes, but than it should be refactored so it will get all the entities as argument and insert them properly updating the primary and foreign keys. 

BTW. why are there no foreign keys and constraints in the runnerup db?

paradix:
As for 3) I got a question.
Renaming Uploader to Synchronizer is one thing, but all the variable names are also often called [prefix]uploader[suffix] 
Also the UploadManager could be renamed to SyncManager, this would make the change even bigger.

Do you want all this to be done?

jonasoreland:
if we *only* do renames in one PR i think that is fine, yes

/Jonas

paradix:
Most points are done. The renaming will follow when all other PRs are accepted.

PR #247, opened 2015-04-27
I10n new ids

bagage:
Hi,

As stated in issue #113, Android seems to support having ID with both time and Time. It is much easier to see what is the expected text when IDs follow English text so I reformatted code to the convention:

English text: A very long text (with stuff) ##and other
ID: A_very_long_text_with_stuff_and_other

Except for array.xml.

It merged, you can close issue accordingly. Do not forget to push on transifex (and translations too!) with `tx push -s -t -f` I think!!

I could not try it because Android Studio complained about unknown command line argument... but it should be mostly fine logically.


jonasoreland:
please check that i did this correctly!!

bagage:
Yes, it's OK. I pushed new ID to transifex too. You can close #113 too.

PR #249, opened 2015-04-29
Add Walking and Orienteering sports

bagage:
I added Orienteering sport with some notes:

- I did clean code a bit. Instead of duplicating sports on multiple places, they are only defined in a few files for now: `Sport.java` for database relation mapping, `values/array.xml` for translation and radio lists, `Constants.java` for database unique ID, and finally in `export/EndomondoSynchronizer.java` or any sportservice for mapping too. Except next point: 
- The last place I did not remove it (but I am thinking about it) is `FeedActivity.java`. I could have used `array.xml` again for sure, but this won't work in any language but English. We should have a verb for each activity (in English it is quite straighforward: bike -> biking, other -> exercing, but not for other languages!). What do you think?
- I did only Endomondo Synchronizer, but it should use "Other" for other synchronizer, right?

### DIFF
--- a/app/latest/java/org/runnerup/tracker/component/TrackerWear.java
+++ b/app/latest/java/org/runnerup/tracker/component/TrackerWear.java
@@ -175,7 +175,7 @@ public class TrackerWear extends DefaultTrackerComponent
                             public void onResult(DataItemBuffer dataItems) {
                                 for (DataItem dataItem : dataItems) {
                                     wearNode = dataItem.getUri().getHost();
-                                    System.err.println("getDataItem => wearNode:" + wearNode);
+                                    Log.e(getName(), "getDataItem => wearNode:" + wearNode);
                                 }
                                 dataItems.release();
                             }
@@ -209,7 +209,7 @@ public class TrackerWear extends DefaultTrackerComponent
     }
 
     private void setTrackerState(TrackerState val) {
-        System.err.println("setTrackerState("+val+")");
+        Log.e(getName(), "setTrackerState("+val+")");
         Bundle b = new Bundle();
         b.putInt(Wear.TrackerState.STATE, val.getValue());
         setData(Wear.Path.TRACKER_STATE, b);
@@ -222,7 +222,7 @@ public class TrackerWear extends DefaultTrackerComponent
                     @Override
                     public void onResult(DataApi.DataItemResult dataItemResult) {
                         if (!dataItemResult.getStatus().isSuccess()) {
-                            System.err.println("TrackerWear: ERROR: failed to putDataItem, " +
+                            Log.e(getName(), "TrackerWear: ERROR: failed to putDataItem, " +
                                     "status code: " + dataItemResult.getStatus().getStatusCode());
                         }
                     }
@@ -370,7 +370,7 @@ public class TrackerWear extends DefaultTrackerComponent
 
     @Override
     public void onMessageReceived(final MessageEvent messageEvent) {
-        System.err.println("onMessageReceived: " + messageEvent);
+        Log.e(getName(), "onMessageReceived: " + messageEvent);
         //note: skip state checking, do that in receiver instead
         if (Wear.Path.MSG_CMD_WORKOUT_PAUSE.contentEquals(messageEvent.getPath())) {
             sendLocalBroadcast(Intents.PAUSE_WORKOUT);
@@ -442,7 +442,7 @@ public class TrackerWear extends DefaultTrackerComponent
     @Override
     public void onDataChanged(final DataEventBuffer dataEvents) {
         for (DataEvent ev : dataEvents) {
-            System.err.println("onDataChanged: " + ev.getDataItem().getUri());
+            Log.e(getName(), "onDataChanged: " + ev.getDataItem().getUri());
             String path = ev.getDataItem().getUri().getPath();
             if (Constants.Wear.Path.WEAR_NODE_ID.contentEquals(path)) {
                 setWearNode(ev);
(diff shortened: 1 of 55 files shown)


jonasoreland:
1) i like many parts, but is scared of the array[] usage.
How about adding Sport.getText(Resources res)
and in there one can validate that dbValue is within bounds of the array,
and otherwise return hardcoded "unknown"

2) there are many sports that one could consider adding...
main reason is that I havent fixed _all_ the uploaders...(syncronizers)

what do you think about that ???
orienteering will with your patch only work for endomondo...

other than that...i think the most asked for sports is walking

/Jonas

bagage:
1) I thought about adding Context in Sport too but it felt wrong. But I think this is indeed the best way of handling it.. will modify.

2) I thought by default synchronizers would use "Other" sport when it is not implemented yet. Is it an issue? I do not have an account on each to test, but maybe I should do that..

And I know that walking is much more asked than orienteering but eh, I am the PR maker :) Obviously the next step would have been to add walking sport, but I wanted to be sure first.

paradix:
If I could add my 2 cents :) Wouldn't it be nice to use the recently implemented Entities when changing existing code. And for instance in HistoryActivity.java use:
'''
int sport = Sport.valueOf(activity.getSport()).getDbValue();

instead of

'''
int sport = Sport.valueOf(cursor.getInt(4)).getDbValue();

bagage:
@paradix what would `cursor` be useful for then? Would it become obsolete or do I need to do something to "load" it into activity object?!

paradix:
You just need to call the proper constructor and you the Entity object instead of the cursor.
```
public ActivityEntity(Cursor c)

bagage:
OK, not a problem. Otherwise @jonasoreland  I am running into different issues where I think you can greatly help me.

- Where did you find Map My Run sports' ID? I did not found them in the only [documentation I found](http://www.mapmyrun.com/api/3.1/?doc)...
- How should RunnerUp Live handle multi sports? Should I let you do this one?
- I learned about TCX format and its sport limitations yesterday and tried, as sugested in #250 to modify workout sport type after upload. You can see my first attempt in 72b2f00; it is based on some code I found (https://github.com/cpfair/tapiriik/blob/master/tapiriik/services/GarminConnect/garminconnect.py line 579) but it is not working... I am not sure aboubt how to do a POST on the given URL without using a session but I think you may lead me to the correct answer here :)
- Otherwise google fit data, endomondo, facebook (limited to some sports only) and funbeat are done and need to be tested.

bagage:
On another side, should not ActivityEntitty be modified? Yet it has `public Integer getSport()` but I strongly suggest to modify it to `public Sport getSport`... but changes should be made everywhere. wdyt?

jonasoreland:
are you waiting for me here ??

jonasoreland:
note: in garmin uploader, you don't use the SET_TYPE_URL, but the upload...which looks wrong.

bagage:
Well, if you can answer my questions it would help for sure. But I still have 6 synchronizers unmodified/todo yet...

jonasoreland:
Hi again,
Sorry for late reply.

1) I like the most parts of the code,
2) I not super with feature being "half-baked"
3) It's hard to "help" you in email/conversation

=>

How about gating the new feature behind an "experimental" checkbox.
This was you, me and other can use/try/work on it...but noob's don't see
it until it's more complete.

What do you think ?

(and yes, there some work of doing that...but not that much,
 maybe we can introduce a generic "enable experimental features")

bagage:
Sure, I agree with you. I will do the experimental setting option so that everyone can work on it since I am stuck for most of the synchronizers ;).

bagage:
I added the  `Experimental features`  checkbox in settings. When enabled it will change sports list in Start tab to add the new ones, but it is not perfect (eg. it uses default Android Spinner row instead of your custom implementation which I did not manage to make it work) - I think this is OK enough since this is only temporary/debug code.

I also fixed Garmin uploader which is now working fine it seems.

I reverted Nike+ because I do not like the API I started to use (it requires data every X secs for the whole run..).

So here is the current status:

Todo:
- TCX uploaders where we should modify workout after upload: 
  - digifit
  - map my run
  - runningahead
  - strava
- nike: [only documentation found is about API](https://developer.nike.com/documentation/api-docs/activity-services/add-activities.html) I tried to use but which looks wrong. 
- runnerup live: no documentation available
- runtastic: the only problem is to find matching Sports ID... at worst we can brut force :)
- jogg.se: no documentation found..

Done and quickly-tested:
- Garmin
- runkeeper
- google fit data
- endomondo
- facebook: known limitation is that it will only support Running/Biking/Walking for ever
- funbeat

jonasoreland:
updated TODO also...does it look correct ?

bagage:
Well, nike / runnerup live / runtastic / jogg.se are missing from TODO list. Can you add them?

jonasoreland:
- nike is only running...i think we should just not upload if it's not running.
what do you think ?

- runnerup live is not that interesting (but I'll add it for completeness)

- i'll add runtastic and jogg

/Jonas

bagage:
Are you sure that nike is only for running? According to [documentation](https://developer.nike.com/documentation/api-docs/activity-services/add-activities.html) it supports more?

jonasoreland:
cool!
1) i had not seen that documentation!

2) but...I have an account on it...(i have one on all runnerup supported
sites)
and I can't find anywhere where it's not about running...

can you ??

/Jonas

bagage:
Hm you are actually right, nothing seems available on the website.. we can skip others sports from being uploaded then!

PR #254, opened 2015-05-13
Download scheduled workouts from Garmin Calendar

CharString:
This will check your Garmin Connect Calendar for the workouts you've scheduled this week.
It prepends the name of the workout with a * if it is scheduled for today.

### DIFF
--- a/app/src/org/runnerup/export/GarminSynchronizer.java
+++ b/app/src/org/runnerup/export/GarminSynchronizer.java
@@ -47,6 +47,7 @@ import java.io.StringWriter;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Calendar;
 import java.util.List;
 
 @TargetApi(Build.VERSION_CODES.FROYO)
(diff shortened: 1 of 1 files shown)


CharString:
@jonasoreland This patch may solve this [bad review](https://play.google.com/store/apps/details?id=org.runnerup&reviewId=Z3A6QU9xcFRPRmF1UWhybGZKWlNfTWU1Q1JXM05CSWEzY0lsWHpEM3RuZ1ZNUDhldVB4cHd4NV9JWXRaWjJCWVllX3F1dFhVanh5b0xwVENDa1otemdEbnc):

>[...] Unfortunately when I attempt to download my workouts from Garmin Connect I get a message stating "Loading workout list (1)", but my workouts don't show and the Download button is grayed out so I am unable to select it. [...]

I think she may have tried to use a Garmin Training Plan. That will create a schedule of workouts and without this patch, those workouts don't show up in RunnerUp.

jonasoreland:
Hi,

I think patch looks pretty good...but I haven't had time to read what it actually does.

Could you please explain a bit more what it does and how it works.
And expand on how it relates to "Garmin Training Plan" that I don't use,and don't know what it is...

Sorry for being lazy...

/Jonas

CharString:
> Sorry for being lazy...

You're excused, you've done quite a lot already ;-)

> And expand on how it relates to "Garmin Training Plan" that I don't use,and don't know what it is...

On Garmin Connect you get a calendar, where you schedule workouts, races etc. They also provide a selection of training plans for several distances and and fitness levels, that prepare you for a race of that distance. You can select, for instance, a 12 week 10k plan of 3 workouts a week, to end on a 10k race you're planning to run. That will fill your calendar with different workouts, a long run, some threshold tempo runs, some intervals. All starting out "short" and gradually building up the distance and intensity, with the occasional taper for extra recovery and an easy taper week before the race.

These workouts, as of yet, don't show up in RunnerUp for download.

What the patch does is before listing the user-defined workouts, construct a url for the calendar based on today's date. Then gets all the schedule items, skips every item that's not a workout (calendar also contains notes, events etc.), then for each workout-item get the workoutId (equivalent to the workoutIds from the workoutlist) and the title (prepended with a '*', if the schedulitem is for today, so you get a visual cue without the need for changing other code) and add those pairs to the list.

I've chosen to do this *before* the user-defined workouts because you either:
  * don't use the schedule (and then nothing changes from before)
  * you *do* use the schedule, and then the scheduled workouts are the items most subject to change, and you would like to present those *first* to the user.


I'll fix the merge conflicts, and then go out for a run. ;-)

Issue #145, opened 2014-12-09
TargetCoaching with HR works poorly with shaky HRM connection


CharString:
For [this run](http://connect.garmin.com/modern/activity/641652018) I set my target heart rate to Heartrate zone 2 (134-147). Given my heartrate is above the maximum of 147, I should get audio feedback like "heartrate 155, slow down". Instead, I got audio feedback like "heartrate 130, speed up". I suspected by feeling that I wasn't going too slow, so I mostly didn't follow up on the feedback with going faster. And sometimes the spurring to go faster would stop.

I suspect that if a [TargetTrigger](https://github.com/jonasoreland/runnerup/blob/master/app/src/org/runnerup/workout/TargetTrigger.java#L24) is used in [CoachFeedback.emit](https://github.com/jonasoreland/runnerup/blob/master/app/src/org/runnerup/workout/feedback/CoachFeedback.java#L84), the 0 valued observations whilst connection is down are added and falsely bring down the moving average.

Maybe a guard in [TargetTrigger.addObservation](https://github.com/jonasoreland/runnerup/blob/master/app/src/org/runnerup/workout/TargetTrigger.java#L132) could help. Something like:

```
if ( dimension == Dimension.HR && val_now == 0 )
    return null;
```

If not... it would get its value indirectly from [GpsTracker.getCurrentHRValue](https://github.com/jonasoreland/runnerup/blob/master/app/src/org/runnerup/gpstracker/GpsTracker.java#L675) which returns nulls that are converted to 0 values, which doesn't really match my audio feedback experience. So for now I think this path is unlikely, but I'm totally new to the code and I don't really do any Java development.

gerhardol:
HR==0 should probably be filtered #477

PR #256, opened 2015-05-14
Fix bug in TargetTrigger.

dibaus:
At the moment, the average is computed over all values in the measure[] array.
However, at the beginning not all values are meaningful, so adjust the computation
to only take meaningful values into account.


CharString:
@dibaus, do you think this is related to #145?

dibaus:
@CharString Yes, in some sense it is. However, my patch won't fix the problem of recording incorrect values due to a missing connection. It will only fix incorrect 'speed up' messages at the beginning of a run (or possibly when a new segment is started). Fixing 'incorrect values is more difficult unless one makes extra assumptions, e.g. that no valid measurement ever can be negative.

PR #259, opened 2015-06-05
Add current cue : instantaneous speed, and current heartbeat rate

ixeft:
This PR add the possibility to have the current speed and heartbeat rate in audio cue


jonasoreland:
looks great!

PR #260, opened 2015-06-07
Add remaining speed audio cues

wbrenna:
Pull request #259 introduced current speed, but speed functionality wasn't yet implemented. This pull request adds total speed, lap speed, and interval speed as well as fixes the cue formatter to output a speed in kilometers/miles per hour.


jonasoreland:
looks great
can you only check for 0 before dividing

/Jonas

wbrenna:
Good catch - added.

PR #262, opened 2015-06-08
AndroidManifest.xml: use sensorPortrait instead of portrait

bagage:
Given my mountain bike phone support and my device which has power button on its size, I need to put in upside down on the support, but currently RunnerUp won't rotate...


bagage:
@jonasoreland bump?

PR #268, opened 2015-06-15
NPE fixed when no start point present

### LINKED ISSUE #239: Crash when opening activity downloaded from runkeeper
I haven't debugged this yet but it happens for 1 in 10 of the downloaded activities when clicking on them from history list

paradix:
fixes #239


jonasoreland:
not entirely sure how that can happen...but sure

PR #269, opened 2015-06-16
start/end point type when downloading runkeeper

paradix:


bagage:
Is not `else if` condition redundant with the `if` one?

```java
+            if (lat == null || lon == null) {
+              // ...
+            } else if (lat != null && lon != null) {
```

paradix:
I'm sorry I'm pretty tired after 2 very long days at work, but how is it redundant?

bagage:
Well if `latt == null || lon = null` is true, then `lat != null && lon != null` is necessary false (and vice versa). So this could only be rewrite using:

```java
+            if (lat == null || lon == null) {
+              // ...
+            } else {
```

paradix:
Yeah you are absolutely right. 

16 cze 2015 22:05 Gautier Pelloux-Prayer <notifications@github.com> napisał(a):

Issue #21, opened 2014-01-27
Use open source map toolkit


jas4711:
RunnerUp currently uses Google Play Services to display the map, but play services isn't free software, which means RunnerUp (with map functionality) can't be distributed via fdroid for example -- they require that the app is only free software.  There exists a bunch of free software friendly map toolkits, and this issue is about tracking progress to add support for at least one of them.

(Of course, and I mention this for completeness, there is the alternative of distributing runnerup without map functionality in fdroid -- but it would be better to have maps there too.)

One solution -- but by all means not the only one -- is mapbox.  I have not used it myself, so I can't recommend it, but superficially it appears to fit the bill.  Please have a look at https://www.mapbox.com/blog/mapbox-android/

There may be other options, maybe the fdroid community can give recommendations.  I'm asking around...

/Simon

patcon:
Unless I misunderstand, I think this can be closed?

PR #272, opened 2015-06-17
map: replace google maps with MapBox/openstreetmap opensource alternative

### LINKED ISSUE #20: Work without Google Play Services?
This app seems to be exactly what I have been looking for -- and I would like to install it on my Samsung S3 which is running Replicany without any GAPPS stuff.  However the app seems to depend on Google Play Services, which is a no-no for me.  Would it be possible to relax this requirement so that it is an optional dependency?  What does the app use Google Play Services for?  Thanks in advance.

Btw, someone mentioned your app on the fdroid forum as well: https://f-droid.org/forums/topic/runnerup-an-open-source-run-tracker-for-android/

/Simon

bagage:
Closes #20 and #21.

Please note that MapBox use `minSdk=9` while we are using `minSdk=8`. I forced mapbox using:
```
<uses-sdk overrideLibrary="com.mapbox.mapboxsdk"
```

But I did not test what happened on SDK 8... another alternative would be to drop SDK8 support?

Apart of that, there are plenty of things to improve (markers icon, route smoothness, etc.), but I think we got here an equivalent of current google maps implementation without its drawbacks (dependency on google play services, etc.). I tried to remove any reference to google maps (including About legal reference).

Last, I currently have set my own key in details.xml for MapBox API access, maybe Jonas you want to use another one? (I just created and account and used default map).


jonasoreland:
looked quickly and it looked great...
1) i will not get to this until i made next release...
2) with android wear additions, this doesn't really remove dependency on
play services any longer, right ?

/Jonas

bagage:
1) no prob
2) where is this dependency used currently? I removed google play from build.gradle and everything compiled fine when building "app". Is google play needed for "wear" build flavor? (which I could not compile even on your master branch because of missing android:elevation key in XML).

skaldarnar:
For me `android:elevation' and `android:stateListAnimator` cannot be resolved, either.

bagage:
Thanks @skaldarnar  for confirmation! We should enter a separate issue for this then. Digging this problem, I got the answer to the item 2) @jonasoreland : I see where wear is actually used... and yes, I will have to keep google play services for latestCompile...
This is not a problem for Fdroid because they do `pre-build` stuff (like removing file etc.), and this can be easily workarounded by doing:

`cp ./app/froyo/java/org/runnerup/tracker/component/TrackerWear.java ./app/latest/java/org/runnerup/tracker/component/TrackerWear.java`

I will update the PR to fix this.

bagage:
Entered compilation issue in #275

bagage:
Any news here @jonasoreland ?

jonasoreland:
Hi,
1) now I looked.

2) I can't help to think that it would be nice to create a new subproject (sort of like hrdevice)
which encapsulates the map...and then inside this one can use either google-maps or the one
provided in this patch...what do you think ? too much effort ?

3) i think the legal notice for play services is needed...when we use play service :)

4) maybe you should make a variant (like "froyo" or "latest") which does not have play services
and can remove all references...

5) how does this work with froyo ?
100 runnerup users still use froyo (2.2)
350 runnerup users still use gingerbread (2.3)

bagage:
3) well, ... yes sure. I am not a licence lover, but is this popup display a Google requirement? Because I do not think it is actually valuable at any point + it causes ANR / crashes (when you try to open it quickly multiple times... even if this is stupid :) ).

5) I did not test but yes, I must find why library minimum sdk is 9. I will investigate this

2) and 4) Well... here are my points:
- When I asked what were the advantages of the Gmaps solution, the only advantage was "it is implemented". Has this changed? Or any drawback to use mapbox maybe? Because otherwise, my next points are null and void.
- Making another variant would mean maintaining another set of file and/or review the whole organisation (for instance a `mapbox` variant cannot build `latest/` folder except `latest/[...]/gmaps` directory). Either have to rename "latest" folder to something generic between gmaps and mapbox variants, or something else... 
- Doing 2 variants also means doing a map interface (to choose from), adding UI stuff dynamically in code (instead of plain XML layout), and limiting potential of this object: anyone wanting to contribute to mapping can do the job a single time, it will not have to do it twice. For instance, I am thinking about nice start/end/pause/resume checkpoint icons which is only possible in mapbox I think - this point makes me think that supporting gmaps is too much effort for almost 0 advantages (... except if you have an answer to 1)).
- Finally, about making another variant, it will be (again) painful: currently latest is not building and we have only 2 variants... 3 variants has to be tested and is more error-prone.. so I dunno.

jonasoreland:
fair enough, lets skip dynamic or build-time choosing of map solution.

repeat:
1) the only advantage to gmaps is that it's implemented.
2) but the only advantage (that i know of) to some-other-map is the "run without play services"...and I already have play services and it also needed for android wear.

anyway...i open to the some other map...but not investigating/asking/pursuing how it should be done seems overly open imo.

/Jonas

ps. why doesn't latest build for you??

bagage:
Yes, one of the reason not to use gmaps was because of "run without play services" - but since this dependency is now only used for variant `wear`, it is OK. If a variant with/without gmaps was available it would have been fine too - but it seems overkill to me.
Advantage of mapbox compared to gmaps are (for me):
-  promotion of openstreetmap 
- mapbox SDK seems far better than gmaps API
- no crash when releasing the non-official version of runner up (eg. its fine if API key is missing) - although we could have fixed it for gmaps but..
- tones of opportunities (multiple rendering available, etc.)
- can cartography the map if some routes are missing after your run.

jonasoreland:
Hi,

How is it going ? Have you found why library minimum sdk is 9?
(this is the only review point currently I think...or?)

/Jonas

bagage:
1) do you want the Google service legal notice back in settings? If this is not mandatory it would be better not to put it, it is not that much useful in my opinion...
2) Well, Froyo has no emulator image and I could not find a device running it at my work today, so I could not test RunnerUp on it... I *think* minimum sdk is 9 due to openGL rendering stuff only available from SDK9, but I asked on [MapBox repository directly to be sure...](https://github.com/mapbox/mapbox-android-sdk/issues/792). It would be great if someone with a Froyo device could test it though...

jonasoreland:
1) i think yes,
(quite sure it is mandatory...and don't feel like doing the research again)

2) hmm...no emulator image, are you sure??

anyway, i think it would be ok for maps not to show on froyo devices...but
NPE doesn't feel right to me.
i.e you should the write code that puts a textbox "not supported on froyo"
instead of the map.
what do you think ?

3) can you give me link to page describing license, account settings/keys
etc for MapBox ?

/Jonas

bagage:
1) done
2) yes, unfortunately :/ 
bonus :) ) I did your suggestion: on Froyo a placeholder text is displayed intsead of the map
3) Basically it is free until 50K users and/or views per month (see [pricing](https://www.mapbox.com/plans/)). Even if Runner Up is reaching this limit, it seems to be feasible to have multiple API keys chosen randomly. By the way it will move the API key to build.gradle so that you can add a secret one for release versions, so that it is not steal/invalidated by spammers making lots of queries.

jonasoreland:
1) don't know if i'm beating a dead horse here,
but just copying the existing xml/code to froyo to make it use google-maps on froyo...
wouldn't be all that hard/bad, or ?

it would mean that maybe DetailActivity.java could be stale in that branch...
but maybe it's better than no map. wdyt ??

details:
2) why did you move AboutPreferences.java, don't understand ?
3) are there color codes instead of e.g "00ffff"
4) i like the handling you made of mapbox.xml

bagage:
1) honestly, as stated earlier handling 2 different maps is annoying if adding new feature on it - I do not think that the 100 Froyo users would mind having it or not. I mean Froyo is pretty old now and almost everyone moved on at least Gingerbread. But that's up to you :)
2) Because on Froyo we do not use google maps at all (we don't have it in build.gradle) so we cannot use `com.google....GooglePlayServicesUtil` in java file so I had to use 2 different versions. Am I wrong?
3) Yes, indeed.
4) Just do not forget to put them in release.properties :)

jonasoreland:
merged (finally?)
thanks for great effort!!

patcon:
I'm new here, but thanks to everyone involved! :fireworks:

bagage:
thanks @jonasoreland for integrating it! You're welcome @patcon (although I did this for me in the first place :) ).

jonasoreland:
FYI this has lead to https://github.com/jonasoreland/runnerup/issues/317
which prevents me from making release and merge other features.
Can you look at it ? (or should I revert back to gmaps ?)

PR #274, opened 2015-06-18
Android 5.0: Change notification visibility to public 

bagage:
Currently when screen is locked, notification content (distance / time / pace) is hidden because its lollipop default behavior. Changing visibility to public set it readable again.
Visibility is only available in API 21 so I bumped build.gradle from API 20 to API 21.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -31,11 +31,11 @@ android {
             minSdkVersion 8
             targetSdkVersion 18
             versionName "1.46"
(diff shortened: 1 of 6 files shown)


jonasoreland:
basically looks good, but please don't create a "customSettings" method.
it's supposed to be version compatibility...not functionality.
i.e make 2 separate methods...(setLocal and setVisibility)

bagage:
split in 3 functions then

PR #286, opened 2015-07-08
Activity summary is now computed automatically in the event of a crash

dhiltonp:


dhiltonp:
I've left the "Recompute" button in the UI for now. I'm that space could be used for a "Merge activity" option.

dhiltonp:
The last activity is checked for the "time" summary field. If not present, the summary is recomputed.

This check is run the first time the History view is opened after starting runnerup.

jonasoreland:
comment: thought about it some more.
if you do STOP activity, and then RESUME
and the phone goes out of battery, this fix won't find it.

to fix that we need a marker (like you suggested)...
my suggestion is that we add a new field, ONGOING
that is set to 1 when activity is created, and cleared
somewhere once it's complete...

but! the current fix will fix most of the occurances i think.
i.e this is maybe not that urgent

dhiltonp:
I'd like to think that looking for the highest location id and then
verifying it was a stop type would work, but I checked my db and I have 50
starts and 40 stops but everything seems to be working fine...

dhiltonp:
It may be best from a maintenance standpoint to just add a flag the next
time we need to change the schema.

jonasoreland:
re add next time: agree

PR #290, opened 2015-07-15
SYNCHRONIZERS: update sports status

bagage:


bagage:
bump here

bagage:
I also have updated various file to use markdown syntax and rich text support.

jonasoreland:
Hi, sorry for delay.

I think you miss interpreted my intentions with the SPORTS field.
I want what sports are supported for uploading.

E.g funbeat does NOT support uploading the sport types that you put in, right ?
it in fact only supports them for the feed...(which it in fact did before adding it to the sports-map)
make sense ?

the other changes look fine...
/Jonas

bagage:
My bad for Funbeat, I updated it. Do you see any other errors???

jonasoreland:
checked briefly...let merge this and we can fix errors

PR #291, opened 2015-07-18
GPS: fix update frequency - currently, we updated position as soon as…

bagage:
… possible by default since we were updating every 0 meters...

### DIFF
--- a/app/res/layout/settings.xml
+++ b/app/res/layout/settings.xml
@@ -224,7 +224,7 @@
             android:title="@string/Poll_interval_ms" />
 
         <org.runnerup.widget.TextPreference
-            android:defaultValue="0"
+            android:defaultValue="5"
             android:inputType="number"
             android:key="@string/pref_pollDistance"
             android:persistent="true"
--- a/app/src/org/runnerup/tracker/component/TrackerGPS.java
+++ b/app/src/org/runnerup/tracker/component/TrackerGPS.java
@@ -82,11 +82,11 @@ public class TrackerGPS extends DefaultTrackerComponent implements TickListener 
             frequency_ms = Integer.valueOf(preferences.getString(context.getString(
                     R.string.pref_pollInterval), "500"));
             if (mWithoutGps == false) {
-                String frequency_meters = preferences.getString(context.getString(
-                        R.string.pref_pollDistance), "5");
+                String freq_meters_pref = preferences.getString(context.getString(R.string.pref_pollDistance), "0");
+                Integer frequency_meters = Math.max(Integer.valueOf(freq_meters_pref), 5);
                 lm.requestLocationUpdates(GPS_PROVIDER,
                         frequency_ms,
-                        Integer.valueOf(frequency_meters),
+                        frequency_meters,
                         tracker);
                 mGpsStatus = new GpsStatus(context);
                 mGpsStatus.start(this);


jonasoreland:
i don't see why we should change from 0 ?
no one has complained about it, it gives good precision.

i think making so that if you bicycle (or walk) the default could be different
would make sense. I.e default value dependant on sport.

also, one could possibly tune in during the activity, but that is much more
work, and questionable if it's worth it.

bagage:
If you don't want to change from 0 then you should set time frequency to 0 ms as well. Sure precision is very "good" since we are updating it 100 time per sec. 
But IMO:
1) this is useless by default (one can still change that in settings )
2) on Endomondo at least my pace is unreadable due to GPS imprecision and update rate . WIth 5m / 500ms settings this is much better and readable.

jonasoreland:
interesting,
i logged into endomondo and checked my last 2-3 runs. it looks quite ok.

what phone/os do you have ?
what is your estimated gps precision ? (my normally show 3m before start).

i checked my db and I get gps updates approximately once per second:

sqlite> select t1.activity_id, avg(t1.time - t2.time) as diff from location
t1, location t2 where t1.activity_id = t2.activity_id and t1._id = t2._id +
1 and t1.activity_id >= (select min(_id) from activity where start_time >=
strftime('%s', '2015-05-01')) group by t1.activity_id;
527|1473.32421340629
528|1002.74382639062
530|998.752079866888
531|1609.44112859468
532|999.615236629473
533|1015.31728665208
534|1024.32126696833
535|1264.74686483976
536|1000.42744176106
537|1366.83037400831
538|999.273519796586
539|1594.50787401575
540|1090.25017535656
541|1000.73313782991
542|1079.8782807151
543|1007.50085236959
544|1011.32075471698
545|1474.23544197738
546|1037.35408560311
547|1333.06623826001
548|1013.25567383009
549|500.0
550|1004.28954423592
551|1026.55771195097

how frequent do you get them ?? (with default settings...and with 5m
instead of 0)

/Jonas

bagage:
OK I misread timestamp in TCX file exported from Endomondo, my bad. Indeed I have only 1 point per second as you, but Endomondo uses a 5 sec refresh rate, which looks better on pace. See the difference for the same run:
screenshot from 2015-07-22 19-10-20

I mean with a default parameters of 500 ms and 0 meter min distance, we are adding points as much as available. For a 15 mph max speed activity, I do not think it makes sense and it makes pace graph less readable than it should. Wdyt?

dhiltonp:
Many sites smooth the data, using an n-window moving average... is that what you're talking about? Or you have verified that data recorded with the endomondo app only has a sample every 5 seconds?

bagage:
I'm comparing the same run uploaded from Endomondo on one device and RunnerUp on another device (so not exact same setup though).
When I export both GPX traces, I can see that RunnerUp has 1 point per sec while Endomondo has 1 point per 8 sec in average.
So if an n-window moving average is used, this is probably done client side for Endomondo. Ultimately RunnerUp track is 8 times more "accurate" than the other one.

jonasoreland:
how about doing moving average when uploading ?

bagage:
I don't know if it's really useful to record so much points in any case, but that's up to you! 

I do not think that doing a moving average when uploading is the right thing to do; it should be either done when saving the activity or never done (inconsistency between Runnerup and synchronizers otherwise).  

The best solution imo is to change default values to something less aggressive; at least I will change that for me.

dhiltonp:
I think it depends on the location and the environment - sometimes GPS
struggles. Perhaps the update frequency could be tied to the gps accuracy?

David

gerhardol:
To high recording rate is why phones are worse than dedicated running GPSes.

GPS on phones are adjusted to give as accurate position in every moment. This will not give the most accurate distance calculation and therefore varying pace/speed. For instance if you stand still, the location will vary. "Running GPS" adjust to to this. RunnerUp (as well as most phone apps) overestimate distances.

I have seen that a GPS point filter will give a better result, both to instance pace and distance. Something like 2s and 20m works well on my Sony Xperia Z3 in open areas, a little to aggressive on trails with many turns. (I have tried this to some controlled measured distances, as well as some I have many Garmin 305 logs for.)

The patch should be reworked and allow setting the time/distance sampling in the GUI. Then the default do not have to be changed.

In the long run, a separate filter is preferred (on top of the filter in the phone firmware). That together with using accelerometer and stepsensor can give better estimates. A complete different function though.

gerhardol:
I often run a trail that is 29.9-30.1km on my Garmin 305, Ambit3 30.0-30.2.
With default settings with RunnerUp on my Sony Z3 in a backpocket, I get 32-36km.
with 2s/10m I get 28.9-29.3 so much better. The Z3 is in the back pocket, not a good position but the track is not corrupted, just jaggy.
There are some 1 star reviews that complains about bad accuracy.

I recommend changing the defaults.
If minimal distance is changed, RU should still update speed/HR etc displayed (this is only done when receiving GPS updates now).

gerhardol:
The PR is for the connect phase, there updates should be frequent.
A similiar change should in my opinion be done for the default logging, a separate PR.

PR #292, opened 2015-07-19
build.gradle: invert productFlavors order so that fdroid detect correct version code

bagage:
See [PR](https://gitlab.com/fdroid/fdroiddata/merge_requests/875/commits) for details but basically fdroid has a tool to automatically retrieve version code from `build.gradle`. However it is really basic and retrieves the first versionCode found, so changing the order will fix the fdroid tool without any drawback on runnerup I guess.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -27,17 +27,17 @@ android {
     }
 
     productFlavors {
-        froyo {
-            minSdkVersion 8
-            targetSdkVersion 18
-            versionName "1.47"
-            versionCode 8000047
-        }
         latest {
             minSdkVersion 14
             targetSdkVersion 21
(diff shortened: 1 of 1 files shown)


jonasoreland:
Btw, i would appreciate if you dropped me a note
when it has been accepted at fdroid

bagage:
Well can you do a new release? So that the first fdroid version is based on a release version and not a random commit?

jonasoreland:
i "just" need to fix mapbox keys and such...but is currently on
vacation...mostly without internet/computer

bagage:
OK, nevermind PR was accepted on fdroid so first build will be available tonight if all went fine :)

PR #296, opened 2015-07-26
Add homescreen widget to display feednews

bagage:
Here's a preview:

feed_widget_preview

Since it has many thing in common with `FeedActivity` I factorized some code.
I think `SyncManager` class is wrong since it does UI stuff (spinner...), but I preferred not to change that to much so I only monkey-patched the class with `if (useSpinner) /*do spinner stuff*/`..

Can you review @jonasoreland ? Thanks!


bagage:
BTW I also improved Nike+ synchronizer a bit by handling user with no friends (test account yee!) and without profile picture (test account yee!).

app/src/org/runnerup/export/NikePlusSynchronizer.java
```diff
@@ -413,24 +409,31 @@ private void getOwnFeed(SimpleDateFormat df, List<ContentValues> result) {
...
-            e.printStackTrace();
-        } catch (IOException e) {
-            e.printStackTrace();
-        } catch (JSONException e) {
-            e.printStackTrace();
+        } catch (Exception e) {
+            if (profile != null) {
+                Log.e(getClass().getSimpleName(), "Failed for profile '" + profile + "' because of the following: " + e.getMessage());
+            } else {
+                e.getStackTrace();
```
jonasoreland: printStackTrace() ?
or rather, why unity the 2, printing null if profile == null

jonasoreland:
only checked briefly (poor internet).
but the "monkey" patching doesnt look nice.

how about a
interface SyncSpinner
{
methods();
};

and that caller supplies one...which can be a empty one for the widget ?

wdyt ?

/Jonas

app/src/org/runnerup/export/NikePlusSynchronizer.java
bagage: Indeed, typo it should be `e.printStackTrace()`, damn Android Studio (and me :)). I don't like printing stack trace in any case, it looks like real crash while it isn't... only printing the message should be enough.

bagage:
Well, I would rather prefer not to do any spinner interface but instead remove spinner code from SyncManager and move it in UI classes, if possible? And since it looks like a big portion of code I wasn't motivated enough to do it... but maybe spinner code interface is the only easy way...

jonasoreland:
if yon can move it out...that would be great...
interface is one of of doing that...

jonasoreland:
any news ?

bagage:
I'll try to do that now, yes (uhh, java...).

bagage:
Done - first commit is minor change in AccountListActivity (only used Android Studio refactoring tool).
Second commit is the progressdialog stub, does it look OK?

jonasoreland:
looked at the 3 latest commit, they looked ok.

shall I look at the entire thing ?

bagage:
Well this is probably not bug-free but the widget code is pretty much straightforward and not very interesting to read... so if you're OK with it, go for merge!

jonasoreland:
i would love to start merging this...but I'm stuck on #317

bagage:
Well please do not merge it yet - I still need to do a few minor things first (do not refresh feed twice...).

bagage:
Hello - I think I am done on this feature @jonasoreland , if you want to merge it please do so.

bagage:
Any news @jonasoreland?

jonasoreland:
forgot, sorry.
but I can't merge now...since it doesn't auto-merge...

bagage:
np, done

app/src/org/runnerup/export/SyncManager.java
```diff
@@ -897,18 +916,20 @@ public void syncActivities(SyncMode mode, Callback synchCallback, String synchro
...
+        if (noButton != null) {
+            noButton.setOnClickListener(
                 new View.OnClickListener() {
                     @Override
                     public void onClick(View v) {
                         synchronized (cancel) {
                             cancel.append('t');
                         }
-                        mSpinner.setMessage(getResources().getString(R.string.Cancellingplease_wait));
+                        if (mSpinner != null)
```
jonasoreland: is this needed ?

app/src/org/runnerup/export/SyncManager.java
```diff
@@ -931,9 +952,12 @@ public void onClick(DialogInterface dialog, int which) {
     }
 
     protected boolean checkCancel(StringBuffer cancel) {
-        synchronized (cancel) {
-            return cancel.length() > 0;
+        if (cancel != null) {
```
jonasoreland: when is this needed ?

jonasoreland:
ok... found 2 things.
i'll merge anyway, can you please comment
or make a new PR fixing them

app/src/org/runnerup/export/SyncManager.java
bagage: yes I think so - from widget you will never have cancel not NULL.

app/src/org/runnerup/export/SyncManager.java
bagage: not at all, fixed thanks.

PR #307, opened 2015-08-11
runtastic: make login work again

darander:
This commit makes it possible to (again) add runtastic accounts.

The issue was that we can't no longer retrieve the authenticity-token at https://www.runtastic.com, instead we need to use https://www.runtastic.com/en/login.

This has been tested and verified using the emulator.


jonasoreland:
super!

PR #309, opened 2015-08-15
Use fragments in MainActivity

daverix:
In order to be able to create a better UI, the first step is to use fragments. This PR will make the app look like this to start with: https://photos.google.com/share/AF1QipMULPT_AQ8vHM6-kmJmzUxCyDCrQaQ89c1XThXjmImpsJGnkXZUD3tfZDgbb6rMgg?key=TTl0Z1Uxa0oteC03XzItTWREYnJvQWpwRkZ4MjRB

I hope upgrading android version target isn't a problem...

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -42,7 +42,7 @@
 	<application
 	    android:icon="@drawable/icon"
 	    android:label="@string/app_name"
-		android:theme="@android:style/Theme.NoTitleBar">
+		android:theme="@style/Theme.AppCompat.NoActionBar">
 
 		<activity
 			android:name=".view.MainLayout"
@@ -62,20 +62,9 @@
 			</intent-filter>
   		</activity>
 
-		<activity android:name=".view.StartActivity"
-		    android:screenOrientation="sensorPortrait" >
-		</activity>
-
-		<activity android:name="org.runnerup.view.FeedActivity"
-		    android:screenOrientation="sensorPortrait" >
-		</activity>
 
 		<activity android:name=".view.RunActivity"
 		    android:configChanges="keyboardHidden|orientation"
-		    android:screenOrientation="sensorPortrait" >
-		</activity>
-
-		<activity android:name=".view.SettingsActivity"
 		    android:screenOrientation="sensorPortrait" >
 		</activity>
 
@@ -85,10 +74,6 @@
 
 		<activity android:name=".view.DetailActivity"
 			android:configChanges="orientation|screenSize|uiMode"
-		    android:screenOrientation="sensorPortrait" >
-		</activity>
-
-		<activity android:name=".view.HistoryActivity"
 		    android:screenOrientation="sensorPortrait" >
 		</activity>
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'android-sdk-manager'
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 21
+    compileSdkVersion 22
     buildToolsVersion '22.0.1'
 
     sourceSets {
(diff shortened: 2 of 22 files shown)


jonasoreland:
i would love to start merging this...but I'm stuck on https://github.com/jonasoreland/runnerup/issues/317

jonasoreland:
Hi,
Can you update this??
now that it looks like v1.49 is finally getting releasable

/Jonas

daverix:
Done :)

jonasoreland:
and...same question about supportfragment (manager)...
is that no longer needed for apps supporting old and new devices ??

daverix:
When using a support fragment, it's LoaderManager is of the support library type so we still use the support fragment manager etc.

jonasoreland:
I don't understand: "Now we still use a froyo version because of some google play things, but if we can solve that too then we only need one flavor."

1) your current PR doesn't break froyo ?

daverix:
No it shouldn't.

jonasoreland:
fyi, was about to merge this...but it had conflicts. will merge manually tonight (or as soon as i find time).

jonasoreland:
tried merging...and found that size of APK increases a lot....specially for froyo
froyo: 1014K => 1.8M (80% increase)
latest: 3.5M => 4.1M (17% increase)

Do you why ??

daverix:
Probably because newer support library have more code and proguard is not
running minimize for support libs?

jonasoreland:
hmm...can you fix so that froyo does not increase as much ?

daverix:
I can take a look in the weekend.

daverix:
Rebased against master, will check apk sizes now.. Is it release you checked?

jonasoreland:
yes, I checked release...
froyo: v1.50 1014K
latest: v1.50 3.5M

daverix:
I have now managed to get down the sizes to the following:

froyo debug: 2.8MB
froyo release: 1.9MB

latest debug: 4.3 MB
latest release: 3.4MB

I think the reason for the larger file sizes corresponds to all new image resources used to get the material design look and feel on devices with version lower than Android 5.0.

jonasoreland:
hmm...froyo is almost twice as big...

i think people on older devices generally like the small footprint...
but does 2Mb vs. 1Mb matter,

wdyt ?

/Jonas

daverix:
I don't know how many still uses old froyo phones but I don't think it
matter much if its 2mb instead of 1mb.

nebmo:
Jonas, I think you have to decide if you want new looks or small app. Can't have both..

jonasoreland:
damn in...should have made these comments prior to merge.
1) missing target annotations! (the ones causing https://github.com/jonasoreland/runnerup/issues/324)
2) the account list in DetailActivity, is super small on my phone. Almost not usable
3) the notes field in DetailActivty is hard to see (it is no longer white)

Can you fix ?

daverix:
1) I didn't cause it, but I could look at it, just not right now...
2) I will go through my changes real quick to see what caused it.
3) Does #328 work for you?

daverix:
2) Does #329 look ok?

jonasoreland:
1) I know you didn't cause that one.
But your patch (that I merged) removed @TargetApi(Build.VERSION_CODES.FROYO) annotations.
And it was *missing* annotations that made us not notice the build failure.
Conclusion: We should have the annotations everywhere, then we get build failure rather
than runtime failures.

2) I commented your other fixes individually.

/Jonas

bagage:
1) concerning that I know it's my fault, but since I never build Froyo personally I do not see them (my fault)..and travis build seems fine with it. 
Do you know any way to detect these invalid API calls from android studio? We could change travis configuration so that it fails in that case.

jonasoreland:
the target annotations makes build fail
if using incorrect apis...

jonasoreland:
pingeling.

how is progress on this? many things are still missing to make this good.
maybe I should rollback and you can fix thing wo/ me harassing ?

/Jonas

jonasoreland:
reverted...see https://github.com/jonasoreland/runnerup/commit/c8cc63f0e0d77d2f9a453714159c9f47ad5ee6a1.

Issue #317, opened 2015-08-31
Failed to make release with Mapbox


jonasoreland:
1) First proguard failed to compile
I added
+-dontwarn okio.**
+-dontwarn com.vividsolutions.jts.awt.**
+-dontwarn android.support.**
+-dontwarn com.squareup.okhttp.**
+-keep class com.squareup.okhttp.** { *; }
+-keep interface com.squareup.okhttp.** { *; }

2) now it seems that the map doesnt work with the new release.
Unclear why...

bagage:
Even if a never used proguard i'll take a look at it until end of week...

jonasoreland:
thx!

bagage:
First thing first, proguard is total sh!t. (seriously, this tool...).
Secondly, this has been broken since #312 (commit 17d01ab) , release or not. Maybe @daverix knows what's going on there?

bagage:
OK fix is easy, i'll do a PR.

jonasoreland:
thx!

bagage:
Concerning the new release, could we just remove the "experimental features" settings which enables extra sports, and just use them always?
Any uploader which does not support them yet will uploads them as "Other" sport, except for Facebook which will drop them (because other not supported). In any case there will still be uploaded to all trackers, right?

jonasoreland:
Hmm...that sounds reasonable.

PR #320, opened 2015-09-04
map_area.xml: mapview must be within LinearLayout to display properly…

bagage:
… (why? black magic is black magic..). Fix proguard in the mean time to solve #317


bagage:
Please do not ask me why a linear layout is needed in map_area.xml while there is already one in top level layout ... it *just* works...

PR #322, opened 2015-09-07
Cleanupsettings

bagage:
As stated in #317: remove "extra sports settings" and use it always instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: android
+
+android:
+    components:
+        - build-tools-22.0.1
+        - android-21
+        - extras
+        - extra-android-m2repository
+        - extra-google-google_play_services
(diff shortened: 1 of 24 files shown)


jonasoreland:
1) I like it
but
2) Can't we keep the experimental features checkbox/setting ?
I think we're going to need it again in the future...

/Jonas

bagage:
2) the thing is ... no-one actually use it if he/she cannot know what the settings will actually enable. I think it's better to add custom setting each time it's needed. I'm not sure many people except me used it until today... wdyt?

jonasoreland:
I'd say keep it... (maybe add the comment for the checkbox that it's
currently unused )
Ok?

bagage:
Added back.

PR #328, opened 2015-09-20
Add spacing around notes edittext to make it more visible

daverix:
device-2015-09-20-220613


jonasoreland:
Hi,
I fact I was thinking about the work-out notes in DetailActivity (i.e when you save).
And...atleast I think it was nice when it was background was white...

Regardless, what ever change we make, we should make it into a style.
(and use it for all occurances...)

wdyt ?

daverix:
My thinking was to eventual redesign using material design themes but I just wanted to take it step by step. :)

jonasoreland:
1) there is no conflict with material design and styles, or ?
2) well, now that this is merged, keep the patches coming then!

PR #329, opened 2015-09-20
Fix row height of share in details activity

daverix:
device-2015-09-20-222217


jonasoreland:
1) I think the actual checkbox is too small. Almost impossible to press (on my 4.5'' phone).
2) Again, whout about putting this into a style and use it for all checkboxes.

PR #347, opened 2015-10-20
Add unit test for AutoPauseTrigger

nebmo:


jonasoreland:
looks great!
comment:
1) should travis to actually run the tests too (i didn't see that it did)

nebmo:
Yeah you are correct! I dont know how though... do you know?

jonasoreland:
no idea...but the travis.yml doesn't look to complicated...so it shouldn't
be that hard

/Jonas

nebmo:
I dont understand, froyo is not building for me:
runnerup\common\src\test\java\org\runnerup\common\util\ValueModelTest.java
Error:(19, 13) error: cannot find symbol class ValueModel
Error:(30, 19) error: package ValueModel does not exist

nebmo:
Changed so it only runs the tests for Latest. Guess that have to be enough for now..  you can merge

jonasoreland:
guess
- when David added the ValueModelTest there where no automatic testing.
- hence they have not been run since
- hence they have rotten

wdyt ?

/Jonas

app/build.gradle
```diff
@@ -61,6 +62,7 @@ android {
         // but continue the build even when errors are found:
         fatal 'NewApi', 'InlineApi'
         disable 'MissingTranslation'
+        abortOnError false
```
jonasoreland: what does this mean ? why is this change needed ?

nebmo: We get lint errors so it would break the build otherwise. Wasn't there another PR that was about lint errors?

nebmo:
I really dont know what is wrong with ValueModelTest. AS gives no clue...

nebmo:
Same problem here I think:
http://stackoverflow.com/questions/28782518/android-unit-testing-not-working-for-android-library-project

jonasoreland:
i rather have you disable the ValueModelTest then add "abortOnError false"

jonasoreland:
(or best would ofcourse be to fix it...)
but disable it is also fine...

nebmo:
Updated build tools and renamed a language folder (copied from other PR). now it works

jonasoreland:
great!

nebmo:
So.. next step is to get some database rows from a workout where autopause didnt work. you have?

PR #350, opened 2015-10-27
Runningfreeonline

jonfre:
My first pull request (ever =)

Added runningfreeonline.com integration (upload account).


jonasoreland:
comments:
1) i will look at it during weekend
2) overall looks good.
3) you should not change version.
4) i'm not happy at all about extra code in SyncManager (for the key),
but haven't really checked (will do in weekend)

jonfre:
4) Yes it sucks, but the user must in this case somehow be informed about the "secret key". I've [requested](http://www.runningfreeonline.com/Talk/g/posts/t/3553) an API change. Maybe generalize a notice for all Synchronizers to implement if needed. To avoid another dialog, any notice could also be included in the username/password dialog.

jonfre:
Status/thoughts?

jonasoreland:
status:
1) i checked, and it seems that your are right.
2) since i have done nothing.

---

do you think it's possible to add a more generic "hint" for the
username/password dialog ?
(instead of the current hard coding)

/Jonas

jonfre:
Ok, I'll figure something out.

jonfre:
Check out the generic notice. I've added a column to accounts table for any string resource id to be shown.

jonasoreland:
LGTM

Issue #298, opened 2015-08-02
No '...' menu in Android Lolipop 5.0


bagage:
This settings menu has disappeared and this can be problematic in various places. Obviously #295 will (have to) fix this but in the mean time we should find a workaround/solution for this...

bagage:
[For reference](http://android-developers.blogspot.fr/2012/01/say-goodbye-to-menu-button.html)

daverix:
It's called overflow menu now and should be put into a toolbar if you want to use it.

tbdal:
on the z1 compact with android 5.1.1 there is no "..." menu and therefore an export of the gpx files is not possible.
is there any workaround yet?

jonasoreland:
not yet :(
sorry.

kas70:
I've found a workaround (at least for cyanogenmod 12 (based on 5.1.1):
You can add the 3-dot menu in the android settings (Buttons->Buttons and layout). 
The 3 dots are now always visible and you are able export gpx-files again;)

jonasoreland:
patch coming ?

kas70:
It's not a change in runnerup, but in the android settings. So no patch available, sorry.

JanKrAppel:
Is there a timeline for a fix on this? I need to have access to the menu in order to export to GPX, and on my device (Android 5.1), the in-app menu button does not show, and the device manufacturer stupidly neglected to provide menu button of any sort.

Alternatively, what about providing either a Dropbox or an ownCloud upload feature in the account settings? That would solve the problem for me, too.

daverix:
There is a fix that can be done in the app... Set target SDK level to 23, add the Android support design library, make the activity extend AppCompatActivity, add a toolbar at the bottom, call setSupportActionbar method and pass the toolbar as a parameter. That way you get a menu button at the bottom right, above the navigation bar (back, home, recent)

JanKrAppel:
I'm not sure what you mean: Are you proposing a fix for the app or are you trying to explain to me how to fix it for myself? Because if it's the latter, I have just about zero knowledge of Android development...

daverix:
I'm proposing a fix for any developer that wants to help. I have just moved
myself so I don't have the setup to do it yet. As soon as I have and no one
else have done anything, I could fix it.

daverix:
I just added the appcompat library for the current target version instead to cause as little other changes as possible. Please review and pull #379  @jonasoreland

jonasoreland:
[...]

PR #379, opened 2016-04-04
Use appcompat to bring back menu button

daverix:
This solves the problem in #298


jonasoreland:
looks great,
haven't tested...(don't have time for another approx. 2 weeks)

Issue #389, opened 2016-05-08
Problem with Runtastic connection


Gittino:
Good morning,
first of all thanks for this amazing app!
I have a problem connecting to Runtastic.
Before I was able to connect to it and upload my data without any problem but from about one week I'm not able anymore. When I try to connect to the account after having entered my username and password it tries to connect but after few seconds it ask me my username and password again.
Could you kindly have a look and at least tell me if there is a workaround to skip this problem? I really would like to upload my new data on the runtastic portal.
Thanks and best regards.

jonasoreland:
Hi,

I have experienced the same actually.
Do you happen to know if it started with the new release, or it also
happened with old release ??

/Jonas

Gittino:
Hi Jonas,
no, it is not a problem of the last release.
I faced it in the previous release also (that should be 1.51 if I don't go wrong).

dhiltonp:
#390 submitted as a potential fix

Bonbadil:
Hello,
I've just installed 1.57 from F-Droid and I get the same problem.

gerhardol:
Runtastic sync is not working and will be hidden/disabled in an upcoming release #590

PR #390, opened 2016-05-09
Runtastic update

dhiltonp:
I used "CharlesProxy" to intercept SSL traffic (vs. static code analysis).

It looks like 2 things have changed for the login - the URL, and a key/value pair has been added.

This code is untested, as I don't have a good android dev setup right now.


dhiltonp:
Should fix #389.

jonasoreland:
Hi,
I merged and tested it. Short answer: it didn't work.
I then reverted the URL change, which made it better. Then it's stuck at getting user id.
(after url2)

dhiltonp:
Ok.

It looks like the "user" key has been changed for "current_user"

    userId = ret.getJSONObject("user").getInt("id");

should be 

    userId = ret.getJSONObject("current_user").getInt("id");

jonasoreland:
i don't get a json document back from the http-request (99% sure)
so there is something broken earlier

/Jonas

dhiltonp:
See email

jonasoreland:
will try find time this weekend...
i might have seen something.

sunfish62:
RunnerUp 1.57 seems to still exhibit this login failure to Runtastic...

jonasoreland:
yea...
i had an idea...but now I have forgotten what it was...
hopefully some runtastic user can pick it up...

jeee:
can't login to Runtastic either.

PR #333, opened 2015-09-28
lint: enable lint checking on travis and fix various errors

bagage:
This fixes #324 and enable lint checking on travis to avoid it in the future.

I also changed cues  plurals (according to lint) so after merging this, you will have to push new sources to Transifex.

The last error which causes Travis to fail is an issue in Transifex probably, I contacted their support and waiting for a reply (within the week) so please do not merge this yet!

PR #400, opened 2016-06-29
Lint cleanup travis

gerhardol:
This is similar to #333 (credit to @bagage) enabling lint in Travis builds

The difference is that 'MissingQuantity','ImpliedQuantity' is not included (as requested by @jonasoreland in #333), to be done later
A number of warnings have been eliminated too (more could be fixed, requires more testing).

Lint in AS and from gradle is not completely synced

Tested on Marschmallow only, but changes should not affect runtime...


jonasoreland:
why have you removed 
> fatal 'NewApi', 'InlineApi'

iirc this is to prevent usage of api's that is not supported by minimum version.

gerhardol:
Fatal anyway,  no need to add explicitly

jonasoreland:
are you sure ??

we had numerous times that one accidentally used incorrect apis before
adding that
but it was a while ago...

/Jonas

gerhardol:
I got newapi errors while eliminating lint warnings, can recheck tonight

gerhardol:
Wrapup for the lintOptions change: 
That will promote the error/warning to fatal when lint is running

NewApi is error by default, so changing to fatal is not much change.
However InlineApi is warning by default and will therefore not result in an error. There are so many warnings that this will not be seen.
I will submit a new patch for this (maybe including more lint eliminations).

PR #401, opened 2016-06-30
Centralize build-tools and other shared versions'

gerhardol:
Simplify updates of versions

Instead of changing build-tools etc in 4 separate files, set it in the common file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,22 @@ language: android
 
 android:
     components:
-        - build-tools-23.0.1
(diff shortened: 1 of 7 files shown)


gerhardol:
Also update build-tools to 24.0.0 (wanted at least 23.0.3, took latest)
The problem was in Travis, platform-tools and tools have to be updated before build-tools
Also caching builds, no speed-up seen though

gerhardol:
..and gradle update....

PR #403, opened 2016-06-30
Lint eliminations

gerhardol:
Reducing more obvious Lint warnings. Few of these are expected to be real problems, they could potentially be hiding real problems though.

A disclaimer is that I have run this very little, only in emulator.
Want to coordinate that with updated SDK level


app/project.properties
```diff
@@ -11,6 +11,6 @@
 #proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
 
 # Project target.
-target=android-18
-android.library.reference.1=../../google-play-services_lib
-android.library.reference.2=../GraphView
+#target=android-18
```
jonasoreland: what does this do/mean?

app/src/org/runnerup/content/ActivityProvider.java
```diff
@@ -155,7 +155,8 @@ public ParcelFileDescriptor openFile(Uri uri, String mode)
                         final boolean includeMap = true;
                         String str = map.export(activityId, includeMap, null).toString();
                         out.second.write(str.getBytes());
-                    } else if (res == RUNKEEPER) {
+                    } else //noinspection ConstantConditions
```
jonasoreland: it possible to instead put the //noinspection somewhere so that the
formatting of the if-statement is "normal".
(no big deal)

app/src/org/runnerup/export/RuntasticSynchronizer.java
```diff
@@ -339,7 +339,7 @@ public Status upload(SQLiteDatabase db, long mID) {
                 fv.put("user_id", userId.toString());
                 SyncHelper.postData(conn, fv);
                 int responseCode = conn.getResponseCode();
-                String amsg = conn.getResponseMessage();
+                //String amsg = conn.getResponseMessage();
```
jonasoreland: how about instead adding it to the logging call ?

app/src/org/runnerup/workout/AutoPauseTrigger.java
```diff
@@ -46,8 +46,8 @@ public boolean onTick(Workout w) {
     private void HandleAutoPause(Workout workout) {
         Location lastLocation = workout.getLastKnownLocation();
         Double currentSpeed = workout.getSpeed(Scope.CURRENT);
-        if (currentSpeed == null)
-            return;
+        //if (currentSpeed == null)
```
jonasoreland: why ?
this can happen, "CURRENT" will return null if no value for 3 second.

hrdevice/src/org/runnerup/hr/Bt20Base.java
```diff
@@ -331,9 +331,7 @@ static BluetoothSocket tryConnect(BtHRBase base, final BluetoothDevice device, i
                 Method m;
                 try {
                     m = device.getClass().getMethod("createInsecureRfcommSocket",
-                            new Class[] {
-                                int.class
-                            });
+                            int.class);
```
jonasoreland: this is not the same is it ?
before: an array with one element which was a int.class
after: an int.class

jonasoreland:
comments:
1) big patch, I think most look good, but not all.
2) general: why not use StringBuilder ?
3) i made lots of individual comments.

maybe easiest split into several commits ?

gerhardol:
A general note is that I followed suggestions from lint/AndroidStudio inspect, if possible automatic.
I choose those that seemed reasonable, did not reformat to keep "consistent".

app/project.properties
gerhardol: This is not needed in AndroidStudio (Inspect)

app/src/org/runnerup/content/ActivityProvider.java
gerhardol: Will try

app/src/org/runnerup/export/RuntasticSynchronizer.java
gerhardol: Will take a look

app/src/org/runnerup/workout/AutoPauseTrigger.java
gerhardol: getSpeed returns "double", so never NULL
Should currentSpeed be a double and check for 0 instead?

hrdevice/src/org/runnerup/hr/Bt20Base.java
gerhardol: Revert
(How could this be suggested and why did I not see it?)

app/src/org/runnerup/content/ActivityProvider.java
gerhardol: No reasonable way, else and if are separate statements
Voted for a bug
https://code.google.com/p/android/issues/detail?id=186172

app/src/org/runnerup/workout/AutoPauseTrigger.java
gerhardol: I added a check similar to what you describe, then the test fails
Reverted the change

PR #404, opened 2016-07-03
Raise SDK to 10 (Gingerbread 2.3.3)

gerhardol:
Opened primarily for discussion, minor file change
Android usage: https://developer.android.com/about/dashboards/index.html
Froyo (SDK 8) has 0.1%, SDK 9 has less than 0.1%. Even SDK 10 has 2.0%
One specific reason to drop Froyo is a change in SDK 23, backwards supported from SDK 9: (see Apache Http Removal)
https://developer.android.com/about/versions/marshmallow/android-6.0-changes.html
While the SDK can be updated still supporting SDK 8, it is another limitation

An alternative to raising the SDK is to drop the "froyo"build completely, just starting with ICS (SDK 14).
If Froyo builds are needed, they can be done from a separate maintenance branch (just few new features supported).
I guess ICS/JB should be supported mainline some time more, for instance Sony Xperia Active is on 4.0.1

There are usage statistics in Play...

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -39,7 +39,7 @@ android {
             versionCode 14000053
         }
         froyo {
-            minSdkVersion 8
+            minSdkVersion 10
             targetSdkVersion 22
             versionName "1.53"
             versionCode 8000053


jonasoreland:
stats for RunnerUp is:
~2% 2.2
~8% 2.3
=> ~10% 2.2-2.3

i still think it's valuable to support these 10%...don't you ?

gerhardol:
The numbers are a reason to try using workarounds to support newer target SDK.
Cancelling this request.
Maybe suggest this again if things get complicated.
At some point it will be easier to backport changes to maintenance branches.

gerhardol:
What about raising lowest "full version" from 14 to 15?
SDK14 was shortlived 4.0, most should have been updated to SDK 15 (4.0.3), not 14
Can you share the full usage statistics (maybe including screen resolution)?

MapBox recent requires SDK 15.
There are some functionality there that are interesting, new maps, performance etc.
(I am specifically interested in WMTS support, could enable Swedish Lantmäteriet Terrain maps).
Upgrade 0.7.6 to 4.x will require some efforts.

jonasoreland:
1) raise to 15, SGTM
2) statistics: attached.
screenshot from 2016-08-01 11 30 08

gerhardol:
SGTM: Sounds Good or Silently Giggling?
http://www.urbandictionary.com/define.php?term=SGTM
(assumes Sounds Good...)

Nice to see the statistics.

PR #409, opened 2016-07-18
Appcompat theme

gerhardol:
Really only 5b70d4d (and later?) to consider
Opened mostly to get feedback


gerhardol:
The button text on pre KitKat is lost when enabling/disabling. Pre-ICS support limits the options.
There seem to be some problems with button text with my master builds too though.

gerhardol:
Replacing with PR #412 

Few changes to this branch but based om master
(Removed singleLine workaround for Workout buttons, not really improving)

PR #412, opened 2016-07-26
Use AppCompat Theme

gerhardol:
Replacing PR #409 

AppCompat theme is required for some newer elements in Support Design library like SnackBars
Using AppCompat also aligns the design with other apps, removes some of the dated look
Some overrides could be removed in code (I did not change the override in styles.xml)

The update slightly changes fonts, dividers and looks, the following "larger" changes are seen (may be more):
 * Backgrounds were previously mostly black, but not all (for instance Advanced Settings were graded). Overrides removed, all gray now
 * Before KK: Button text in manage Workouts have visual glitches when enabling/disabling, issues before too though.
 * Before KK:  CheckBoxes have slightly different appearance than stock and before

I have tried to use singleLine and various other fixes for buttons but do not find a solution that works without glitches.
I propose that visual glitches pre KK are accepted.

Note: AppCompat v22.2.1 seem to miss some functionality compared to v23. Maybe easier to improve visuals with the related SDK-24 PR too (the SDK-24 will require this PR).
Anyway, AppCompat is required to freshen up the GUI to be more standard AndroidManifest

Longer term, the app design should be updated.
Maybe Light theme?
Use more Material Design, skipping the two levels of tabs and use a action bar?

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -42,7 +42,7 @@
 	<application
 	    android:icon="@drawable/icon"
 	    android:label="@string/app_name"
-		android:theme="@android:style/Theme.NoTitleBar">
+		android:theme="@style/AppTheme">
 
 		<activity
 			android:name=".view.MainLayout"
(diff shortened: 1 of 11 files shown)


gerhardol:
Screwed up base branch, resubmitting as #413

PR #408, opened 2016-07-17
Sdk version update to 24

gerhardol:
Few API changes from 23->24, so going with 24
Currently no specific needed from SDK 23, but there eventually will be

Most changes are due to the updated permission handling in SDK 23

Note that the commit contains other branches too (Swedish, lint). I hope that those are included first. The specific commits are from 3, 10, 17 July. (I have only really tested them together.)
However, this is partly for discussion, expect to resubmit after reviews.

PR #414, opened 2016-07-26
Code changes to support requesting MM permissions dynamically

gerhardol:
GPS is requested at App startup. If not granted a SnackBar will nag the user (unless not ask again is ticked). (This is tricky to do when pressing Start button, the request is done asynchronously.)
A few corrections to handle GPS not started
Longer term, the GPS can be set to work like other sensors, to be able to deactivate it

Also for read/write backups. There it would be natural to always request permission, but this do not work, so popup asking the user to manually activate.

Only popup for storage permissions
The request is not working there for some reason. (This could be improved with at least supplying a link to the settings).
This should be good enough for now.

Note that sharing activities and workouts uses "internal" directories that do not require permissions.

Note: CircledImageView.setForeground() listed as added in SDK 23
Compile possible with compile SDK 22 though, seem strange that this occurs. However, few Wear should use SDK22 anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,15 @@
 language: android
 
+jdk:
+  - oraclejdk8
+
 android:
     components:
         - tools
         - tools #update to latest, cannot control version
         - platform-tools #latest
         - build-tools-24.0.1
-        - android-22
+        - android-24
         - extra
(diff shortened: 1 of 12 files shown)


gerhardol:
Replacing PR #408

PR #415, opened 2016-07-26
Sdk 24 http update

gerhardol:
SDK depreciates org.apache.http.HttpStatus

The use of HttpStatus error codes is incorrect mix of the modules. These changes should be safe after a code review. RU uses HttpURLConnection  except in RunnerUp live. RU Live requires some verification, I have not been able to do that, is there a server?

This is not a required merge, I submit it to finalize the change I started with before finding the workaround.
Several possible outcomes of the PR:
  * Accepted as-is
   - RU Live can be tested somewhere?
   - RU Live is considered obsolete
  * Accepted except for RU Live (also keep build.gradle then). Obviously, a new PR is needed
 * Ignored, the branch is kept as a reference (at least in my clone)

Note: This is based of SDK-24 but could be applied to SDK-22 as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,15 @@
 language: android
 
+jdk:
+  - oraclejdk8
+
 android:
     components:
         - tools
         - tools #update to latest, cannot control version
         - platform-tools #latest
         - build-tools-24.0.1
-        - android-22
+        - android-24
         - extra
         - extra-android-m2repository
         - extra-google-google_play_services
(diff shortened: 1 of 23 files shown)


jonasoreland:
i don't think there is a RU Live server alive anymore.
Weide added features year ago...he had a server, but it's probably down now.
There has been efforts to make that code more generic, but I don't think it has succeeded.

Btw, both facebook and runkeeper (and probably others too) has "Live" functionality...

PR #420, opened 2016-07-29
File synchronizer

gerhardol:
"Account" to save tcx/gpx to local file similar to webaccounts
Simplifies handling of "unsupported" accounts, in my case a PC application
I use FolderSync to manually transfer the files to my PC, this eliminates the fiddling with sharing the activities

Based on share-activity-update, but not related
While changing the files, I also cleaned up Lint
Copyright on the file added set to me, can be changed


jonasoreland:
build failure ?

gerhardol:
Suppression also for command line lint

Why is AS Inspect separate from Lint? I would like to use the same configuration.
This check is not configurable in Inspect, it is always on, but it is smarter: ThE version is checked before the call.

Still, I should have checked the private Travis build.

jonasoreland:
Is it possible to (somehow) move the askFileUri into FileSyncronizer
(instead of having it in SyncManager)

gerhardol:
Not in a simple way. The check needs activity etc for the popup.

To move askFileUrl into askUsernamePassword() is possible, but the code gets very complex (I did that first). The code could be moved, but then callbacks etc will have to be configured. The moving would be 'forced'.

FILEPERMISSION is an AuthMethod parallel to OAUTH2, USER_PASS so it is natural to handle it the same as the other methods (even if only used by one account it is easier to add more file formats in FileSync than to create a new Sync).

FILEPERMISSION uses some more of the config fields for accounts. (For instance FORMAT is not really used). The existing AuthMethods only needs to communicate AUTH_CONFIG back and forward, FileSymc needs slightly other info. I have used a middle way with auth_config still being used as a token with information but only FORMAT being used additionally.

jonasoreland:
ok

gerhardol:
Some thoughts after last message. Really nothing related to what you asked, but maybe cleaner.

One change that could be done that maybe cleans up:
 * Include the full config in Synchronizer.upload(SQLiteDatabase db, long mID, ContentValues config) (overloading of course). The FORMAT etc is then used more naturally. It is possible to use the database too to extract data.
 * Always just include AUTH_CONFIG to Synchronizer.init() (This function as well as connect() is used in various phases, hard to handle additional parameters. I added FORMAT too here) Alternatively always read all db data before calling init().
 * Remove unused database fields: DESCRIPTION, AUTH_METHOD (and FORMAT, that could be included in AUTH_CONFIG for FileSyncronizer) (AUTH_METHOD need to be checked though).

PR #429, opened 2016-08-04
Graphview update

gerhardol:
Note: While this works fine for me, I currently suggest this is merged after next update.
The PR #428 should be merged first (there is a merge conflict for version number though).

Update, basically same functionality
Added zoom/scroll as well as 'tap-to-get-value' (as Toast)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -104,9 +104,7 @@ dependencies {
     compile('com.mapbox.mapboxsdk:mapbox-android-sdk:0.7.4@aar') {
         transitive = true
     }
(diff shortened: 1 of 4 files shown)


gerhardol:
I could build fine on my machine, also using the maven repo builds. Strange.

However it does not run in the emulator, Froyo crashes, the library has minSDK 9. 

Disabling starting GraphView handles that.
I still propose that Froyo is dropped. (Easier to make special changes in a separate branch.)

gerhardol:
This PR disables graphs also on GingerBread
The proper fix is in PR  #431

gerhardol:
Handled in #436

PR #431, opened 2016-08-06
Mapbox 4.1.1 sdk15

gerhardol:
Update to latest MapBox
See separate commits for more info

The APK size increases from 3.3 to 14,5 MB, the Froyo decreases to 2,7MB
The "Froyo" release should be set as a separate application like RunnerUp simple or something.
GraphView should probably be dropped there too.

Also handles PR #429 correctly


jonasoreland:
3.3Mb to 14.5Mb ?
is it worth it ???

gerhardol:
For recent phones, I definitely think so. Maps look much better and adds other possibilities.
Few phones the last years have had less than 8GB. Just Play Services is now 40MB so 10 MB is not much.

However, I suggest the Froyo build is released as a separate application to give users a choice.
Name "RunnerUp (NoMap)" or "RunnerUp - Classic"?

jonasoreland:
hmm...let someone else do that...
i just intended to merge...but there are conflicts (didn't check what)

gerhardol:
I will submit a new PR with combined #431 and #429 support, plus some cleanup (target some time next week, will be away for the weekend)

gerhardol:
Handled in #436

PR #435, opened 2016-08-30
Antplus update

gerhardol:
Update ANT+ to latest version
Probably not critical
Also use libs/ folder instead of submodule (ant+) or dynamic download (Samsung)


jonasoreland:
i'm actually not that keen on having libs in the repo.
i prefer both the submodule and the download.

why do you want them in the repo ?

gerhardol:
submodules are a way to share common sources. I find submodules quite difficult top work with (I use it extensively at work), merges are hard and updates are easily lost.
 
Binary dependencies should not be stored in git, it increases repo size. Maven repo is a better way to handle such dependencies with an id and a version. None of these dependencies are in maven repos though.

The Samsung lib is not version controlled at all right now and depends on that the remote url (in this case latest in github repo) is not changed. In this case, cloning the original repo, then adding that as a submodule is an option. However, I propose the artifact (library) is handled the same as for antplus. 

The antplus subrepo contains a lot of non related binaries, the only interesting is the .jar file.

I find no reason to use submodules here, rather use just one repo.
The reason I see is to keep the main repo size down, keeping binaries as "someone elses problem" in the subrepo. The contents of the submodule should then only be the library.

gerhardol:
New PR #442 only updating ANT repo

Issue #300, opened 2015-08-05
Sync to runalyze


piratenpanda:
I don't know if runalyze can import sync data but as it's also free software it would be nice if you could cooperate in some way :)

https://runalyze.com/login.php

sporkbyspork:
This would be really great. By self-hosting Runalyze, you wouldn't need to give the data collected by Runnerup away to third parties.

jonasoreland:
hmm...it seems this is also open source,
https://github.com/Runalyze/Runalyze
so it should be very doable to fix synchronization.

lenucksi:
That would be splendid for the reasons @sporkbyspork named!
They appear to already support quite a number of sync services.
:+1:

mipapo:
Should be no problem, when we have our API. Please stay tuned. We will inform you!

RunningCl0ud:
I cross my fingers and hope you make it happen @mipapo  :+1:

powellc:
Anyone else working on this? Otherwise I've got a pull request coming in a few here.

mipapo:
@powellc Without our API it will be just a dirty hack. So my recommendation is to wait for it ;)

powellc:
Yeah, after I posted that I took a look at Runalyze source and realized what I was up against! Also I haven't done PHP work in years so I don't trust myself to work on that codebase ... I'll wait. In the meantime, just getting gpx files in ownCloud would be awesome, so I'm working on that.

Marx2:
I've succesfully launched selfhosted runalyze site, and imported all my trainings from RunnerUp through gpx files (downloaded from Strava). It would be great to have it automated. And it would be great to integrate live positioning on the map during run with this webapp.

Marx2:
As their API doesn't seem to appear in a near feature, maybe it would be nice to implement so called "dirty solution" 
referenced here:
https://github.com/Runalyze/Runalyze/issues/498
https://gist.github.com/mipapo/cafe45b0c2a247d9c0614bd2cc8fe4d0

mipapo:
I'm not sure if it would be a great idea to do this. 
I really try to get the basic API done by the end of the year.
Next RUNALYZE release will be 3.0 and I think we will just getting faster after that version in our development process. I'm getting more confident that we manage that this year

rickyepoderi:
Hi everyone,

[...]

PR #437, opened 2016-09-06
Syn to Runalyze (Issue #300)

rickyepoderi:
Issue #300 : Sync to runalyze.


Marx2:
Great! I hope it will be in next version

rickyepoderi:
I have just pushed my repo. One change was added, the "type" of the activity is blank now. I have changed my mind and now no type is sent (just running). You can later edit the activity in runalyze and assign the type you want.

There is an error in the integration (but I don't know ehat), please comment me any change.

jonasoreland:
travis failed...
looked at it briefly, don't understand why.

rickyepoderi:
Ok, if you see something wrong just say it and I'll do the changes (I have little experience with android and the annoying android-studio).

VCleemput:
Hey everyone,
I'm completely new to this, so sorry if I'm being to intrusive.
I was wondering, if it would be possible to do send the type of sport, since runalyze doesn't calculate VDOT if no trainingtype is specified.
Thanks!

rickyepoderi:
In runalyze there are two fields to categorize the sport (sport and type) but in runnerup there is only one. At first I did two different methods to pass both parameters (the methods are still there):

* sport: [runnerupSport2RunalyzeSport](https://github.com/rickyepoderi/runnerup/blob/master/app/src/org/runnerup/export/format/RunalyzePost.java#L295)
* type: [runnerupSport2RunalyzeType](https://github.com/rickyepoderi/runnerup/blob/master/app/src/org/runnerup/export/format/RunalyzePost.java#L327)

The problem is we have to map the values in code (with the problems of ids and so on) and besides for the type we have nothing. If you see the second method just passes always 1 which is _jogging_ for _running_ and nothing for the rest of sports. That is why I decided to remove this field when I realized it was not compulsory, and don't pass anything for the type. You need to manually set this field later after uploading to the activity in runalyze. I don't know, if you think is better to pass something (_jogging_ or whatever) it is just reverting this commit: https://github.com/jonasoreland/runnerup/pull/437/commits/8056ed51173edbc3e4df1b0cfac960d8681b415c

One off-topic question, I usually use f-droid in my phone and runnerup is not being updated for a long time (I'm using my own compiled app)... Anyone knows who is in charge of uploading this app to f-droid? I want to go back to f-droid version. :wink:

VCleemput:
Aha, I knew there would probably be a reason :)
Also thanks for the commit-number. I'm going to try and see what it's like.

Marx2:
Can't we add an option for mapping sport type? It even can be a magic number of Running sport type, because it's set once and will not change. 
It will be much more convenient to read somehow this number and set it in properties of account, than changing sport type  every time new activity is uploaded.

rickyepoderi:
IMHO adding a configuration for the type is a bit strange, think that the account screen is shared to all account backends (garmin, runalyze, strava, and so on).

Remember that runalyze uses IDs for sport and type (a number which is the PK in the database) and in my straight forward integration I'm just using default IDs created by the default installation (I mean that if you start deleting and adding sports and types in the server you can screw the integration up). The correct way is having an API (restful or whatever) to discover sports and types defined in runalyze and try to find a more intelligent strategy to map them. The guys in runalyze know this problem and in a future version the API will be provided. Until that time I think we have to keep the integration as simple as possible, so if you think that the type should be sent I'd just choose one default ID (jogging, tempo run or whatever) and fill the type with it.

Open a new issue with the problem and we can talk and decide something.

Issue #434, opened 2016-08-19
Map and Graph tabs incorrect


jluttine:
I'm using kilometers as my units, not sure if it's related to that. Total time, distance and pace are correct. However, the distance markers in Map tab and pace&distance in Graph tab are somehow incorrect. My total distance was 7.90km but Graph shows 9.53km and computes pace values also incorrectly. I switched to Miles units but these are still wrong. Any ideas?

AlwaysBiking:
Same here.  No solution yet :-(.

gerhardol:
The distance in the graph is calculated from data saved in the database and may give different values.
Dont know why this occurs (it for sure is an issue when playing around in the emulator, should not be this obvious in real devices) but the solution I have suggested to @jonasoreland  is to store distance in the database. However, this requires some migration when updating, first time starting will be slow. See what Jonas wants.

PR #441, opened 2016-09-07
Precision was set to float before storing in SQLite

gerhardol:
SQLite REAL stores as double but double values from the GPS sensor were converted to float before storing.
This was clearly visible in the graphs, especially if playing around in Genymotion: The reported total distance did not match the graph distance

This was reported in issue #434 too


jonasoreland:
great digging!

PR #452, opened 2016-10-01
Fixed a bug created by a recent change that broke wearable support.

aaroncyoder:
This commit introduced a bug that broke wearables:
[Refresh Graph handling](https://github.com/jonasoreland/runnerup/commit/8023ced7a1f483f25887c9ea6c954279cad59afa)

Specifically, the change in TrackerWear.java that moved the isGooglePlayServicesAvailable() check to a helper function had two problems:
1 - the logical test was inverted (!= instead of ==)
2 - the (this.)context variable was not initialized before the new helper function was called

This change restored wearable functionality in my own testing.  Please consider merging this pull request or a suitable alternative.

Thanks!

### DIFF
--- a/app/latest/java/org/runnerup/tracker/component/TrackerWear.java
+++ b/app/latest/java/org/runnerup/tracker/component/TrackerWear.java
@@ -103,16 +103,18 @@ public class TrackerWear extends DefaultTrackerComponent
     }
 
     private boolean hasPlay() {
-        boolean res = false;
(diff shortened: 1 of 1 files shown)


gerhardol:
Looks good to me (that introduced the bug, no Wear to test with)

There is a potential similar problem related GoogleApi with "context" in AboutPreference (not occurring now though). Separate PR

PR #453, opened 2016-10-01
Activate sport types walking, orienteering for most languages

gerhardol:
Transifex needs to be updated too
If "sportEntriesExperimental" was not translated, the standard appears still.
Also remove "Not_much_to_say_at_this_point", no longer used (warning in some situations)
Facebook activity type set to Biking only for Biking (was biking for all other than Running before. Could have set Other to Biking but most of the code handles Other as Running)


ataliba:
Amazing :) I need this to use the portuguese language on my phone :+1:

gerhardol:
Any comments on this?

Other Transifex are required too
I can make a cleanup if I get admin access at Transifex

jonasoreland:
can you request transifex access ?

gerhardol:
I have transfer as a user

PR #458, opened 2016-10-03
Manual distance picker miles

gerhardol:
Manual add distance picker mixed miles and meters

Also set defaults

Still presented as m without unit, but that is a general setup
TitleSpinner: mValue should be split in the presentation and value.

### DIFF
--- a/app/res/layout/start_manual.xml
+++ b/app/res/layout/start_manual.xml
@@ -50,6 +50,7 @@
         android:layout_below="@+id/manual_date"
         android:prompt="@string/Type"
         android:text="@string/Start_hour"
+        android:defaultValue="now"
         runnerup:type="timepicker" />
 
     <org.runnerup.widget.TitleSpinner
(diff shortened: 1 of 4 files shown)


jonasoreland:
the usage of 1000 has to be wrong ?

gerhardol:
The change is to instead of entering meters for the minor unit is to enter thousands. So if you set 1.002 miles this would be 1612m.
An alternative would be to let the picker just set complete km/miles and parse the string.

gerhardol:
Any comment on this?

gerhardol:
Any suggestion how to not mix miles and meter?
One field with miles/km where the advance affects complete km/miles?

gerhardol:
Suggestion: This should be reworked to have one field only but +/- steps complete km/miles

magnacustos:
Any update on this. Because it shows as m instead of mi when I upload to runalyze it throws off the distance.

gerhardol:
My pull request was rejected, I still consider this as a bug though. How should it look like?

gerhardol:
Superseded by #624

Issue #440, opened 2016-09-07
Dropdown lists don't stay open


alonevino:
In every dropdown menu/list I see the options just a fraction of second and then they disappear without giving me the chance to select one of them 
example:  start -> basic, tap on "audio cue settings"; it's possible to see (just see) the two options (default and 100m) without having the possibility to select one of them because the list closes

ajhepple:
I have seen this issue also, on my Daughter's phone but not on my own. I can provide system details and app versions if this would be helpful.

RunningCl0ud:
Got that problem to on my SG S5. I think it begun with the last version of RunnerUp.

Marx2:
Same for me. It's almost impossible to choose different training type

jonasoreland:
yes, we need details on where this happens...

jonasoreland:
it's most certainly some upgrade of support libraries...
but it will be hard to fix unless we can reproduce...and I don't have it on
any on my phones :(
@gerhardol any idea ?

ajhepple:
Version 1.53 on Sony Experia with Android 4.4.2 **does not** have this issue. Version 1.54 on Motorola G with Android  5.1 **does** have this issue.

There is a notable difference in the appearance and style of option lists in each case. The earlier RunnerUp version has what I would describe as a modal radio list that fills most of the screen, whereas the problem version is a small drop-down menu style that appears for a fraction of a second then disappears. 

Let me know if you need any other system information.

Screenshot of problem menu: 
![issue 440](https://cloud.githubusercontent.com/assets/11554087/18815504/c68d8634-832b-11e6-8349-1a2bd49ccd55.png)

gerhardol:
The emulators and my Sony Z3 with 6.0.1 Concept (AOSP) works fine.
My workphone Samsung A3 2016 with 5.1 (stock) has a similar problem. The time is occasional short, it is like pressing slightly longer has a longer timeout so no problem selecting.
Sport in Manual is OK, not sport in Basic. The handling is slightly different internally for basicSport (no explicit init) and manualSport.
For Audio Cue Setting, manual mode timeout is slighly shorter than in interval/advanced and they seem to be initialized the same.

gerhardol:
This seems to occur when GPS is on.
If I deactivate Autostart GPS (my normal setting) this is OK until GPS is started. At least in the emulator and on my Z3.

RunningCl0ud:
[...]

PR #459, opened 2016-10-17
Droplist gps update

gerhardol:
Attempt to at least improve issue #440 
I dont like it, but probably better than nothing...


gerhardol:
I have not found a better way to solve this, believe it should be merged

gerhardol:
A better solution in #476 , closing

PR #462, opened 2016-10-31
Back in StartActivity for non Autostart GPS should stop gpsBoundState

gerhardol:
There is currently no way to stop the GPS if you have started GPS but do not start an actvity.
You have to remove RunnerUp from recent apps.

This change stops GPS with Back button, without exiting the app
(Home, recent apps have no effect as before).


gerhardol:
Any comment on this?

PR #471, opened 2016-11-20
Runneruplive apiaddress setting

gerhardol:


jonasoreland:
didn't the runneralizer patch add the possibility of adding generic strings to an account setting ?
can't you use that (ranther than hard code for live)

gerhardol:
The ui still need to handle the generic strings so it still need to be
hardcoded.
(similar for File there you have to remove and add to change formats. OK
for me)

Migrate storage is a second step.

Den 20 nov. 2016 08:13 skrev "Jonas Oreland" <notifications@github.com>:

gerhardol:
Removed the setting

There is no natural place to put the API address. FORMAT could be used (The Java name should be renamed then) as this is only actually only used by File. (URL is generally used, something like AUTH_CONFIG could be specially handled too, that is internal to the account.)

Suggest no change right now, maybe add a comment to AccountActivity.java.

jonasoreland:
not super happy, hope to see refactoring of this.

gerhardol:
I agree it is not nice, but better than having a stray settings in the general settings
Dont know how to refactor in a general way

Issue #394, opened 2016-06-06
[Feature request] Lower playing music during audio cues


pace6666:
Add Audio option to lower playing music during audio cues

Marx2:
With PlayerFM it's stream is not interrupted, and RunnerUp has much higher sound level

pace6666:
With my lg g4 and spotify I have to pause music option otherwise it's not clear

alonevino:
the feature doesn't work for me

gerhardol:
Option in Audio preferences, may not have been working in all situations, see #475 
This should be included in v1.58

davidedelvento:
This is also related to #604

PR #475, opened 2016-11-23
Store prefs_mute as boolean, show it as checkbox

johnlepikhin:
Current implementation is very complicated and furthermore it simply doesn't work in non-English locales (at least in locales where "yes" is written with different characters).

### DIFF
--- a/app/res/values/pref_keys.xml
+++ b/app/res/values/pref_keys.xml
@@ -25,6 +25,7 @@
 	<string name="pref_pollInterval">pref_pollInterval</string>
 	<string name="pref_pollDistance">pref_pollDistance</string>
 	<string name="pref_mute">pref_mute</string>
+	<string name="pref_mute_bool">pref_mute_bool</string>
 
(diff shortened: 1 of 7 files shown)


johnlepikhin:
Probably related issue #394 (some people reported that feature doesn't work for them)

jonasoreland:
maybe paranoid, bur how about changing the value of pref_mute to a new string,
and use the old value verbatim when converting. this to minimize risk that getBoolean is
used on yes/no string ?

johnlepikhin:
Done. Note, you must add resource string/pref_mute_bool. I have no idea how to add it.

jonasoreland:
to add a new "pref_key" you edit (these are not translated)
app/res/values/pref_keys.xml

to add a new string you edit (these are translated)
common/src/main/res/values/strings.xml

it's that easy...and then someone else will do translation for you...

now you can update #478 too, right ?

johnlepikhin:
Thanks, done.

PR #476, opened 2016-11-23
Another try to fix jonasoreland/runnerup#440

johnlepikhin:
Another try to fix jonasoreland/runnerup#440

Show all TitleSpinners as modal dialog. This type of spinners cannot be closed with parent update. Another point - it is bigger and thus more usable outdoor :)

### DIFF
--- a/app/res/layout/title_spinner.xml
+++ b/app/res/layout/title_spinner.xml
@@ -46,6 +46,7 @@
 
     <Spinner
         android:id="@+id/spinner"
+        android:spinnerMode="dialog"
         android:layout_width="0dp"
         android:layout_height="0dp" />
 </LinearLayout>


gerhardol:
A better solution than #459 , should be used instead

johnlepikhin:
@gerhardol What buttons do you mean?

jonasoreland:
@gerhardol will you drop #459 then ?

alonevino:
Can we have the compiled app with this mod? I can also volunteer to be a beta-tester, just let me know. Thanks

jonasoreland:
i just merged it...it will be in next beta-relase.
hopefully this weekend
are you part of runnerup beta testers ?

alonevino:
Thanks for the good news and...no, I'm not part of the beta-testers (so far): what I have to do to be part of the group?

jonasoreland:
it it a google+ group, called runnerup beta testers (or something like that
if you join that, you'll get the beta releases autmatically.

/Jonas

gerhardol:
Note: This change is only active in Ice Cream Sandwich and above.
Froyo and Gingerbread will have to deactivate Autostart GPS

PR #477, opened 2016-11-25
hrValue==0 means nothing

johnlepikhin:
There are two reasons to remove this check:

 1. Some HRM devices reports hrValue==0 for some time, right after it restored contact with the body. This state causes reportConnectFailed(), which starts new full cycle of reconnection, while RetryingHRProviderProxy actually is just successfully finished reconnection.
 2. reportConnectFailed() is called from onConnectionStateChanged(), so it is useless here.

In worst case, very first retrieved hrValue is zero in a series of reconnections => all reconnections will be failed until attempts limit is reached. No HR recording until application restart!

### DIFF
--- a/hrdevice/src/org/runnerup/hr/AndroidBLEHRProvider.java
+++ b/hrdevice/src/org/runnerup/hr/AndroidBLEHRProvider.java
@@ -176,17 +176,9 @@ public class AndroidBLEHRProvider extends BtHRBase implements HRProvider {
                             BluetoothGattCharacteristic.FORMAT_UINT8, 1);
                 }
 
-                if (hrValue == 0) {
-                    if (mIsConnecting) {
-                        reportConnectFailed("got hrValue = 0 => reportConnectFailed");
-                        return;
-                    }
-                    log("got hrValue == 0 => disconnecting");
-                    reportDisconnected();
-                    return;
+                if (hrValue > 0) {
+                    hrTimestamp = System.currentTimeMillis();
                 }
-
-                hrTimestamp = System.currentTimeMillis();
 
                 if (mIsConnecting) {
                     reportConnected(true);


johnlepikhin:
Tested it on workout. Now HRM works much better.

Looks like this patch also fixes another issue: sometimes HRM sends HR=0, it doesn't mean it is disconnected. HRM just wanted to say it didn't detect any heart rate for the last period. Without this patch RunnerUp starts reconnection cycle.
I checked logs, it sent zero twice while 1 hour workout. Each time after a few seconds valid value was received.

jonasoreland:
which HRM do you have ?
I must have added the code for a reason
- darn that I didn't comment...maybe it's in the git logs ??

I have a polar h7 ble

/Jonas

johnlepikhin:
Mine is HRM is some NoName brand from Aliexpress :) Anyway, it works great with all trackers.

This code was added with initial commit of this file:

commit 5cf3302ab648b39a5205f38643656023c1639279
Author: Jonas Oreland <jonas.oreland@gmail.com>
Date:   Sun Sep 8 18:18:08 2013 +0200

    Add SamsungBLEHRProvider

johnlepikhin:
Sadly, there's no information about this case in specification. But for me it's clear: onConnectionStateChange() is called when connection state is changed and current HR value means nothing.
I also played a bit with my HRM and found that it's very simple device with no support of Sensor Contact feature (see https://www.bluetooth.com/specifications/gatt/viewer?attributeXmlFile=org.bluetooth.characteristic.heart_rate_measurement.xml ). So, probably it just sends HR=0 when lost contact and has no any other way to say about this.

gerhardol:
The current behavior has some extra safety for devices that "want to be restarted" with hr==0.
I believe that the proposed patch should be accepted though. Maybe add option "ignore hr==0".

gerhardol:
This was added in c82f9dfda7 a quite big commit, cannot say where it is needed
Having in option is hard for the user
Maybe disconnect after a timeout of 60s or so?

I have a crappy hr belt that acts similar (I normally dont use it).

gerhardol:
Proposal: Change the behavor to restart hr after one minute or so of zeros

gerhardol:
Closing, merging #826

PR #479, opened 2016-11-25
Add HRM connection audio feedbacks

johnlepikhin:
Second try. Merged with master.


jonasoreland:
Overall: very good!
But some comments:
1) isn't there a "hrm-enabled" method or so that you can use to disable the trigger
when the workout is running...the WorkoutBuilder can't check that...

2) what is initial value...i think it should be silent if everything works...

3) how about a hysteresis, so that it waits a few sec before declaring HRM ok again
(if it's flipping)

jonasoreland:
4. Adding similar functionality to GPS signal would be great...

johnlepikhin:
1. If HRM is not connected, HR value will never changed from zero to any other value. Thus, the trigger will not ever cause audio cue. NB: I'm not sure I understand you correctly.

2. Good point, bugfix commited.

3. Well, I tried to add a counter into the trigger. It works well when HRM just loses contact to body for few seconds and continues sending zeroes. But it breaks trigger when real disconnection happens, because HRM backend sends zero only once...

4. Yes, initially I was thinking about GPS audio cues :) But in my case much more actual feature is this one (it is not true anymore since I commited patch for hrValue==0).

PR #482, opened 2016-11-26
Translation refresh

gerhardol:
Enable Norwegian, Catalan, Indonesian, Lithaunian as cue languages

Lint
Transifex update: nb_No added, ru major, ca refresh
Manually added the new languages to the nb/ca/lt/id array file
Set lint 'InconsistentArrays' as fatal, this is required for step_intensity_types for example. Ignore for languages. There is no check that cueLanguageCues/cueLanguageNames matches. This handling should be rewritten.
Add new cue languages also for Swedish


jonasoreland:
LGTM.
Just checking that you're following translation procedure.
1) pull latest translations with tx pull -a
2) commit new translations in git
3) push new strings with tx push
4) make PR for new translations

PR #483, opened 2016-11-26
Apply geoid altitude correction to WGS84 altitude (EGM96)

gerhardol:
GPS elevation is reported to the Geoid, not above sea level, at least in most GPS. (For instance GENYmotion reports sea level altitude though.)
Reported elevation can be off several hundred meters.

GPS elevation is replaced with map elevation normally services, not just the offset but the reliability is poor. Sea level elevation is primarily needed if elevation is presented on the device (realtime or in graph) or used as to calibrate the barometer elevation (to be added).

There are several models for the geoid, egm96 is within 1/10 m for the reference points, better than GPS elevation accuracy.
The database is quite big. I have adopted an implementation to compress the data (both storage and RAM), basis is here https://github.com/gerhardol/EGM96/
Still adds 0.8MB to apk size, so only added to Latest build


jonasoreland:
just checking,
what is LICENSE of this package ?

gerhardol:
MIT, so it can be included in GPL.

PR #490, opened 2016-12-04
Android sensors

gerhardol:
Support internal sensors for step(cadence), temperature, pressure(elevation)
This also needed the infrastructure for those measures

The commits are split to explain some of the internal parts, but most changes are in one of the commits


jonasoreland:
very cool!
1) i agree that this will have to wait until i made release
2) can you rebase it ? i see lots of stuff that I already merged, and patch
looks big anyway.

gerhardol:
The PR is based on the second to last commit in master, actually on PR #486 (uses some of those changes).
(The master update was done just before I pushed the branch.)
So a rebase would only remove one commit (and require a new PR).
The original branch had some hundred+ commits, rebase/squash was manual to comment/document the changes.

PR #496, opened 2016-12-15
Transifex bs

gerhardol:
Transifex Bosnian translation

A few languages added for resConfig too (AppCompat languages, saves some apk bytes). I do not know where these strings may be displayed though


gerhardol:
In #498

PR #500, opened 2016-12-17
Use the appropriate parameters for requesting location updates

mwleeds:
My phone died about 5 miles into a run and I suspected that RunnerUp was using too much battery. This patch should help. Thanks for the app by the way!

### DIFF
--- a/app/src/org/runnerup/tracker/GpsStatus.java
+++ b/app/src/org/runnerup/tracker/GpsStatus.java
@@ -20,6 +20,7 @@ package org.runnerup.tracker;
 import android.Manifest;
 import android.annotation.TargetApi;
 import android.content.Context;
+import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
 import android.location.GpsSatellite;
 import android.location.Location;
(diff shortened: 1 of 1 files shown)


jonasoreland:
the Gps status class is not used when running.
it's only used before the activity is started...so this patch can't affect
battery usage during run.
furthermore, I think it's good to get gps asap in before the activity is started.

summary: the patch as such is good, but I don't think it should be merged.
wdyt ?

gerhardol:
I do not believe this should be changed, it may slow down GPS aquisition in situations where you want lower consumption.

This is only used when connecting.
Decreasing location updates while running will decrease the power if you set a quite high value probably at least tens of seconds to see a difference. So you can use RU in sparse recording situations like a hike where detailed recording is not so important.
I believe the default setting (500ms, 0m) should be increased to at least (2s, 5m) to increase accuracy (less jaggy path). That will not affect the battery though. (A filter that drops points in the application based on the accuracy could be used too, maybe not better but easier to market to those that believe that the more saved points, the better).

mwleeds:
> the Gps status class is not used when running.

didn't realize that. I'll close the PR

> I believe the default setting (500ms, 0m) should be increased to at least (2s, 5m) to increase accuracy (less jaggy path).

Increasing those values would make location updates less frequent, which would make the path more jaggy, unless I'm missing something.

> That will not affect the battery though.

I think it would, unless another application is requesting more frequent location updates.

PR #508, opened 2016-12-18
WorkoutSerializer: Suspicious check for null

gerhardol:
When eliminating Lint warnings, this issue came up
As a change is needed, a separate PR is opened

Not sure at all what need to be done here

### DIFF
--- a/app/src/org/runnerup/workout/WorkoutSerializer.java
+++ b/app/src/org/runnerup/workout/WorkoutSerializer.java
@@ -434,10 +434,8 @@ public class WorkoutSerializer {
                 break;
             }
             case RESTING:
-                if (convertRestToRecovery && duration.first == Dimension.DISTANCE &&
+                if ((!convertRestToRecovery || duration.first != Dimension.DISTANCE) &&
                         duration.second != null) {
-                    // fall through
-                } else {
                     js.step = Step.createPauseStep(duration.first, duration.second);
                     break;
                 }


jonasoreland:
why do you want to do that ??
1) it looks 100% wrong.
2) the construct is imo much harder to read/understand

gerhardol:
Is the current code really correct?
Assign if null?


Den 20 dec. 2016 09:23 skrev "Jonas Oreland" <notifications@github.com>:

jonasoreland:
what do you mean ?
i don't see any assignment.
which assignment do you mean ?

gerhardol:
In the current version, createPauseStep is called if duration.second == null

case RESTING:
- if (convertRestToRecovery && duration.first == Dimension.DISTANCE &&
+ if ((!convertRestToRecovery || duration.first != Dimension.DISTANCE) &&
duration.second != null) {
- // fall through
- } else {
js.step = Step.createPauseStep(duration.first, duration.second);
break;
}

Den 20 dec. 2016 12:04 skrev "Jonas Oreland" <notifications@github.com>:

jonasoreland:
yes,
and otherwise it's fall-though (as commented)


/Jonas

gerhardol:
The original code:
                if (convertRestToRecovery && duration.first == Dimension.DISTANCE &&
                        duration.second != null) {
                    // fall through
                } else {
                    js.step = Step.createPauseStep(duration.first, duration.second);
                    break;
                }

Original code is a little hard to follow, there is a lint on it.
Rewritten as the following:
                if (!convertRestToRecovery || duration.first != Dimension.DISTANCE ||
                        duration.second == null) {
                    js.step = Step.createPauseStep(duration.first, duration.second);
                    break;
                }

duration.second is used even though it is null
if this is the intended behaviour, the PR should be closed.

jonasoreland:
it is intended behaviour...duration.second *can* be null, it does not have to be...
i think the rewrite is much harder to follow...
this is the "verbal interpretation"
for a STEP of type RESTING
if convertRestToRecovery and duration=distance duration value = 1km
  convert to recovery
else
  create a pause step

the difference between the 2 is that paus is not recorded.
the "duration.second != null" means that RECOVERY should be "finite"
i.e one does not have to press "NEXT" to exit it...

PR #509, opened 2016-12-19
Lint: extractNativeLibs to reduce app size downloaded/stored

gerhardol:
For 6.0 and later, Play can extract and exclude unneeded native libs. (The compress strategies are better too, unrelated here)
This should reduce the MapBox size requirements

Can only be tested with Play...

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -19,11 +19,7 @@
 	package="org.runnerup"
 	xmlns:tools="http://schemas.android.com/tools">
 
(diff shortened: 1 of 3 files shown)


jonasoreland:
why is the min-sdk version removed ?

gerhardol:
The version in gradle is used anyway this is just confusing. There is a
lint for this

Target sdk was already removed from one of the manifests

jonasoreland:
hmm...so this should have been removed when it was added to gradle....
ok.

PR #518, opened 2016-12-30
Lint: Performance: Overdraw: Painting regions more than once

gerhardol:
Note: Unable to test. If problem, the theme should be updated

If rejected, the issue should be suppressed (in the .xml files or lint.xml specifying the .xml files)


jonasoreland:
lint is nice,
but testing is much better. i won't merge if it's not tested.
do you have more PR that are not tested ?

in general, i think there is too much lint attention...isn't there ?
i see lots of PR, apk is only getting bigger and bigger, but few new features !?!

gerhardol:
To get help from Lint either just use Fatal (as done now in RU) or try to eliminate all. I have done a push for the latter trying eliminate all Lints. The Lints that really should change anything are separated (there has been maybe 25 PR due to this). Remaining changes in #524.

I have no Wear HW and problems with the emulator so no tests there right now. Will look into to this again whe #524 is submitted (even if that will be rebased and submitted again after review).

When Lint is done, it will be easier to add new features. But the GUI should be changed first and to continue having Froyo/Gingerbread support in the main branch is a blocker.

PR #521, opened 2016-12-30
Lint: Wear enable lint

gerhardol:
Also in Travis
The app:lintLatestRelease done in Travis is from the app: perspective.
wear:assembleRelease is not really needed though in Travis (but do not affect timing)


jonasoreland:
can you fix this ?

PR #527, opened 2016-12-31
Transifex and Android Lint does not agree on plurals for Polish

gerhardol:
Lint want one,few,many,other, Transifex do not accept "many".
Transifex seem to be correct, cannot suppress Transifex anyway
http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html?id=l10n/pluralforms

Adopting to Transifex, suppress lint error. Handpatching in Transifex sync is a pain.
Similar to Turkish

Puzzled: Why did Turkish appear some weeks ago and Polish now?

### DIFF
--- a/app/lint.xml
+++ b/app/lint.xml
@@ -14,6 +14,7 @@
     <issue id="MissingQuantity">
         <!-- Transifex refuses quantity="one" for tr, (gradle) lint requires it, Inspect allows both-->
         <warning path="res/values-tr/cues.xml" />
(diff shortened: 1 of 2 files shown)


gerhardol:
Seems like Transifex has updated for pl again, latest pull agrees with this commit
The explicit suppression may not be needed in the long run (keep for now)
   <ignore path="res/values-pl/cues.xml" />

PR #572, opened 2017-06-21
Prepare for Android Studio 3: Add Google repo, set flavorDimension

gerhardol:
Also update Android gradle plugin
The preparation has no functional change for AS2.3, but it easier to switch between the versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,11 @@ before_cache:
 
 android:
     components:
-        - tools #latest for "builtin" sdk (24.4.1 in Android-24)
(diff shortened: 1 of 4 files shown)


jonasoreland:
shall I do this before releasing 1.57 ? (or 1.58...now that you made 2 fixes)
WDYT ?

gerhardol:
Does not matter when this is merged.
The change is for development, probably easiest to release first, then
merge.
The latest fixes are not important in my view, but I do not use distance
feedback...

2017-06-22 18:24 GMT+02:00 Jonas Oreland <notifications@github.com>:

Issue #567, opened 2017-06-14
Add JSON Workout to app?


sunfish62:
I am a new user of RunnerUp, and I very much like what I see. 

I generally like to run interval sessions, but need a little more configurability than I see on the Interval tab, specifically in regard to the warmup and cooldown. These currently are hard coded into the app as continue until tap steps. I much prefer to have timed warmup and cooldown, but have no way to change them on the Interval tab.

I thought that perhaps I could add my own custom workouts in the advanced tab. Currently, there are three workouts listed there, and digging around here shows me that these workouts are individual Garmin Connect JSON files that are stored here in runnerup/app/assets/bundled/app_workouts.

Is there a way to add other JSON files directly to my phone so that they appear in the options? Or do I have to add the file to the sources and build the app locally? I will note that I am not a programmer, and have zero experience with java, so building the app will be a challenge.

gerhardol:
You can create and export workouts from Settings-Workouts-Manage eWorkouts
No way to import what I know (now when Garmin import is gone, should be hidden in the app).
The "asset" is included in the app itself, so no way to change that right now. Workaouts are internally stored in a SQLite database.

There should be an option to create workouts from the tab, interval and advanced/interval should be merged.
It should be possible to import .json from a file

sunfish62:
Thank you for your replies, both here and on my other threads. I really appreciate it. 

I agree that creating workouts should be accessible from the tabs.

I asked about importing json files because I didn't see how to create workouts in RunnerUp, or how to edit the workout in the Interval tab. I figured out how on the Garmin connect site to make a workout and save it locally as a json file. I figured that I might then import a workout json file. This functionality is not really required, given the existence of the Create Workout option, which creates a json entry on the Advanced tab anyway.

To discuss the tab issue a little more, I think the Interval tab could be eliminated altogether. The Advanced tab could be renamed Workouts, and a default Interval workout be added to the initial set of offerings in app_workouts.

gerhardol:
The GUI should be refreshed at RunnerUp with a normal toolbar and navigation drawer before other changes are done.
I started that briefly some months ago but await pull requests to be merged before I continue.
I also would to prefer to focus development on new Android versions only, maintaining older versions separately.

jonasoreland:
you can import json workouts (from garmin)
by mailing them with correct mime-type and
then runnerup will import then...

the easiest way of seeing how that work is by "sharing" a workout,
that will send such a mail...

[...]

PR #583, opened 2017-07-22
Hide json extension from workouts

davidedelvento:
@jonasoreland Here is my first PR, as mentioned in #567 

I made the change as small as possible, so it should be almost trivial to review (besides a change in `common/src/main/res/values/strings.xml` which could have been its own PR, but that seemed really overkill for just one byte of a description string....)

Note that this is my first android contribution ever, and first Java project in about 10 years -- I wish there were some unit tests to make sure I did not break anything else. In particular, I have no way to try anything in the Garmin section, and I'm not sure my changes might have affected that. So please somebody checks that before merging.

### DIFF
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -280,7 +280,7 @@
   <string name="Create_new_workout">Create new workout</string>
   <string name="Set_workout_name">Set workout name</string>
   <string name="Downloading_1s_will_overwrite_2_workout_with_same_name">Downloading %1$s will overwrite %2$s workout with same name</string>
-  <string name="Delete_workout">Delete workout </string>
+  <string name="Delete_workout">Delete workout\n</string>
   <string name="RunnerUp_workout">RunnerUp workout: </string>
   <string name="HinHere_is_a_workout_I_think_you_might_like">Hi\nHere is a workout I think you might like.</string>
   <string name="Share_workout">Share workout…</string>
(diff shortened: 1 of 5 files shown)


common/src/main/res/values/strings.xml
```diff
@@ -280,7 +280,7 @@
   <string name="Create_new_workout">Create new workout</string>
   <string name="Set_workout_name">Set workout name</string>
   <string name="Downloading_1s_will_overwrite_2_workout_with_same_name">Downloading %1$s will overwrite %2$s workout with same name</string>
-  <string name="Delete_workout">Delete workout </string>
+  <string name="Delete_workout">Delete workout\n</string>
```
gerhardol: Should be applied to all translations, to be fed back to Transifex after the merge

davidedelvento: @gerhardol  So are you suggesting that I do that change? Or that I take it out of this PR since then it will not be just one byte anymore? Or are you hoping that the translators see this comment and apply it to the language they're responsible for? Thanks!

gerhardol: I recommend that you drop this part, such formatting is normally better done in code.

If you make this change, make the same in all languages. 

@jonasoreland also needs to make sure that GitHub and Transifex is synced. I have opened #574 for syncing from Transifex, after this PR is merged a sync back to Transifex is needed too (without changes in between). Changes in two systems requires timing.

davidedelvento: Understood, thanks for taking the time to explain it to me. I updated the PR accordingly.

PR #590, opened 2017-07-29
disable garmin sync

gerhardol:
Some sync/feed/workouts are not working:
 * Garmin (new API to login, $5000)
 * RunnerUpLive (no server?)
 * Runtastic (reportedly not working)
 * GoogleFit (WebView cannot be used but separate APIs to connect)
 * RunningFreeOnline (site seem to be down)

It is possible to configure the hidden/disabled accounts from the menu. Long pressing the items will toggle enable/disable.
It is not possible to enable/disable just part of the functionality now (like disabling feed but not upload).

Note 1: This PR builds on #589. The PRs may be reviewed separately but should likely be merged simultaneously.

Note 2: The database version is not updated so existing installation is not disabling the accounts. This is intentional ( a separate PR will follow), to be able to move back and forward among branches more easily.


gerhardol:
MapMyRun should be disabled too.
Will not add that to this PR, before the merge order is known.

PR #593, opened 2017-07-31
account remove url

gerhardol:
Prepare to remove FORMAT from Account to minimize changes in the PR to actually change the db.

Building on #589 and #590, only two unique commits

### DIFF
--- a/app/latest/java/org/runnerup/feedwidget/FeedWidgetService.java
+++ b/app/latest/java/org/runnerup/feedwidget/FeedWidgetService.java
@@ -135,35 +135,6 @@ public class FeedWidgetService extends RemoteViewsService {
                     } else {
                         rv.setViewVisibility(R.id.feed_widget_item_notes, View.GONE);
                     }
-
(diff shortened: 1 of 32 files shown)


davidedelvento:
It's not clear to me what the purpose of this change is, can you please elaborate? And I mean the goal, the why, not the mechanics. Note: this is probably due to me not using any online service myself, that "sharing thing" could be completely gone from RunnerUp and I would not notice as long as I keep getting my GPX/TCX file in my local directory.

jonasoreland:
I also don't see any real point with this...
changes to DB are painful since they means that you can't (easily) rollback an upgrade.
i think it's better to leave fields unused to avoid this problem.

what is the feature lurking behind these changes ?

gerhardol:
See #594 for more information, more information there (I was interrupted and could not create the PR directly).
This PR prepares for a DB cleanup. 

One unique commit (and one Runalyze cleanup). Clearly marking the parts that should not be specially handled.

PR #594, opened 2017-07-31
Remove DB.ACCOUNTS unused columns

gerhardol:
The ACCOUNT table contains several columns that are no longer used
These are confusing (the FileSynchronizer was unnecessary complicated due to this).
All "account specific data" is now in AUTH_CONFIG.

Most preparations done in #593 (with preparations in #589 #590 ).
The actual database change has been kept minimal (to simplify moving back and forward between branches).

### DIFF
--- a/app/latest/java/org/runnerup/feedwidget/FeedWidgetService.java
+++ b/app/latest/java/org/runnerup/feedwidget/FeedWidgetService.java
@@ -135,35 +135,6 @@ public class FeedWidgetService extends RemoteViewsService {
                     } else {
                         rv.setViewVisibility(R.id.feed_widget_item_notes, View.GONE);
                     }
-
-                    // c.put(FEED.FEED_TYPE, FEED.FEED_TYPE_ACTIVITY);
-                    // c.put(FEED.FEED_SUBTYPE,
-                    // getTrainingType(o.getInt("TrainingTypeID"),
-                    // o.getString("TrainingTypeName")));
-                    // c.put(FEED.FEED_TYPE_STRING,
-                    // o.getString("TrainingTypeName"));
-                    // c.put(FEED.START_TIME,
-                    // parseDateTime(o.getString("DateTime")));
-                    // if (!o.isNull("Distance"))
-                    // c.put(FEED.DISTANCE, 1000 * o.getDouble("Distance"));
-                    // if (!o.isNull("Duration"))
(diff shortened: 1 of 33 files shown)


gerhardol:
Summary of changes

#589
Addressed the issue with R.* in the database, forcing me to nuke the database at updates

#590 
Building on #589 Disabling the accounts that do not sync

#593
Building on #590 Prepare for a DB cleanup

#594 
Remove the unused columns, changes kept minimal

gerhardol:
@jonasoreland Any comments on accounts not affecting DB version?
Could be a separate PR, but discussing it here so there is only one database version change.

I propose B) (database change).

A) Hack solution
A simple solution is to insert accounts everytime accounts are configured (as well as configure workouts) view are opened.
This Adds 50ms (in emulator) the first time configure accounts is opened. (The time can likely be decreased with optimized SQL.)

+ Simple to implement, snippet below
- Adds UI delay, hack solution

Other: Instead of the static variable, the status could be stored in Preferences so it survives restarts, but that has complications.

//the static variable should be moved to DBHelper, just concept
//DBHelper.insertAccounts() must be changed to static too

     static Boolean dbAccountsAreInserted = false;

    @Override
    public Loader<Cursor> onCreateLoader(int arg0, Bundle arg1) {
        if(!dbAccountsAreInserted) {
            DBHelper.insertAccounts(mDB);
            dbAccountsAreInserted = true;
        }

B) Add "Account version" to new table
Run insertAccounts at startup if the "account" version number is changed. 
Just a check is much faster (<1ms for 1000 checks in the emulator) than insert, so it can be done at startup (or when opening the Account List, this is more "hacky" though, should be combined with the static variable).

+ Relative simple
- Adds time at every startup

C) Only keep configured accounts in db
Change structure, not use CursorAdapter and query the modules.

+ "Proper" solution
-- Bigger rewrite for minor benefits

D) Temporary and virtual tables
Only add configured accounts to persistent DB, add all accounts to a temporary table and access the combined information using a virtual table.

+ Simple to implement
+ Slightly cleaner than a pure B) solution
- Adds time at the startup

E) Combination of B) and D)

gerhardol:
Minor update (columns were not deleted properly)
See also #596 for an implementation of alternative B) above

gerhardol:
Replaced with #611

PR #596, opened 2017-08-02
account content version

gerhardol:
Keep a separate "ACCOUNT VERSION" variable, to keep most account updates out of the DBVERSION

Follow up to #594 , only last commit unique
The content version is checked when the database is opened, which is once until the app is moved out of memory (onDestroy())
This add <1 ms at the startup when version is changed and <5ms when version is updated (in emulator), should be acceptable.

### DIFF
--- a/app/latest/java/org/runnerup/feedwidget/FeedWidgetService.java
+++ b/app/latest/java/org/runnerup/feedwidget/FeedWidgetService.java
@@ -135,35 +135,6 @@ public class FeedWidgetService extends RemoteViewsService {
                     } else {
                         rv.setViewVisibility(R.id.feed_widget_item_notes, View.GONE);
                     }
-
-                    // c.put(FEED.FEED_TYPE, FEED.FEED_TYPE_ACTIVITY);
-                    // c.put(FEED.FEED_SUBTYPE,
-                    // getTrainingType(o.getInt("TrainingTypeID"),
-                    // o.getString("TrainingTypeName")));
-                    // c.put(FEED.FEED_TYPE_STRING,
-                    // o.getString("TrainingTypeName"));
-                    // c.put(FEED.START_TIME,
-                    // parseDateTime(o.getString("DateTime")));
-                    // if (!o.isNull("Distance"))
-                    // c.put(FEED.DISTANCE, 1000 * o.getDouble("Distance"));
-                    // if (!o.isNull("Duration"))
-                    // c.put(FEED.DURATION,
-                    // getDuration(o.getJSONObject("Duration")));
-                    // if (!o.isNull("PersonID"))
-                    // c.put(FEED.USER_ID, o.getInt("PersonID"));
-                    // if (!o.isNull("Firstname"))
-                    // c.put(FEED.USER_FIRST_NAME, o.getString("Firstname"));
-                    // if (!o.isNull("Lastname"))
-                    // c.put(FEED.USER_LAST_NAME, o.getString("Lastname"));
-                    // if (!o.isNull("PictureURL"))
-                    // c.put(FEED.USER_IMAGE_URL,
-                    // o.getString("PictureURL").replace("~/",
-                    // "http://www.funbeat.se/"));
-                    // if (!o.isNull("Description"))
-                    // c.put(FEED.NOTES, o.getString("Description"));
-                    // c.put(FEED.URL,
-                    // "http://www.funbeat.se/training/show.aspx?TrainingID="
-                    // + Long.toString(o.getLong("ID")));
                 } else {
                     Log.e(getClass().getSimpleName(), "Unexpected feed type");
                 }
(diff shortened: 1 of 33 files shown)


davidedelvento:
Watching this @davidedelvento

gerhardol:
@davidedelvento 
Any review comments?

This issue (including #594) updates the database version which makes it harder to move back and forward, even if future updates are simpler.

Should this be merged early or late?

davidedelvento:
@gerhardol I did not study all the details of this PR, but I did study the ones relevant to the `onUpgrade` call, which I believe is the most critical. 

My first comment is that database code needs to be crystal clear, with no edge cases and possible misunderstanding. Unfortunately the starting code is not like that. So I recommend that before this PR, a clean up happens. Here are some comments in that regard:

* [line 239](https://github.com/jonasoreland/runnerup/blob/a873ef379ececf71a84cf7523380a9f95355b97a/app/src/org/runnerup/db/DBHelper.java#L239): this should never happen. The infrastructure should call [`onDowngrade`][1] in this situation. Also, it's not clear why downgrades are not supported since all the code after seems to make backward compatible changes. See also [this][2].

* line [270](https://github.com/jonasoreland/runnerup/blob/a873ef379ececf71a84cf7523380a9f95355b97a/app/src/org/runnerup/db/DBHelper.java#L270): the line is too long and very hard to scroll in the github interface. 
 
* in general, best practices are not followed, e.g. [backing up the tables or add things if not exist](https://stackoverflow.com/a/3505944/25891). Because [`onUpgrade`][3] will automatically roll back the changes if anything goes wrong, and things may exist if we ever implement downgrades (and I think we should)

* all the `if`s in the form of `oldVersion > 0 && oldVersion < 5 && newVersion >= 5` are very confusing. In fact `newVersion` should always automatically be strictly greater than `oldVersion`, and I don't see why the history of the versions should ever had used a negative number (the oldest commit that I could find, namely ae2735d43a49e6afec19381af83f8e9c256d30d3 used DBVERSION=4)

I will get to comments for your PR in a moment.

  [1]: https://developer.android.com/reference/android/database/sqlite/SQLiteOpenHelper.html#onDowngrade%28android.database.sqlite.SQLiteDatabase,%20int,%20int%29
  [2]: 
https://stackoverflow.com/questions/43200408/cant-downgrade-database-from-version-n-to-n-1-on-samsung
  [3]: https://developer.android.com/reference/android/database/sqlite/SQLiteOpenHelper.html#onUpgrade%28android.database.sqlite.SQLiteDatabase,%20int,%20int%29

davidedelvento:
Here are my comments more specific to your PR:

* line [257-266](https://github.com/jonasoreland/runnerup/pull/596/files#diff-205358cc6f1916dbfcd170d78f7283e3L257): not clear to me why have you commented this out 

* line [346 and subsequent](https://github.com/jonasoreland/runnerup/pull/596/files#diff-205358cc6f1916dbfcd170d78f7283e3R346) you could have merged the two large `if`s (the other one starting in line [325](https://github.com/jonasoreland/runnerup/pull/596/files#diff-205358cc6f1916dbfcd170d78f7283e3R325) and made it clearer), but yet I still don't see this `oldVersion` being less-equal to zero

So maybe I'm missing something substantial? If so, further comments from me are moot, so let me stop here for now

gerhardol:
@davidedelvento 
Existing setup:
I am picky too (look at the cleanup PR...) but have not tried to clean up the existing db handling except those I have been directly working with. That is a separate PR (can I assume you will review such a PR?).

onUpgrade() populates data in tables too, why it is called in onCreate(), DBVERSION can be 0. Table manipulations are not needed if creating new db. (This should be changed.)
The transaction structure seem OK to me: If the upgrade script is not matching the structure, onUpgrade() should make a rollback.
onDowngrade() is very hard to implement, it puts constraints on what can be changed...

Updates:
line 257: The account table is recreated for DBVERSION 31->32, so not needed.
346: I separated the handling of ACCOUNT_VERSION and ACCOUNT to be clearer (if the compiler cannot optimize the extra if, we can live with that).

However, I will cancel #594, #596 and make two new PRs:
 * The first is basically 594+596 but will not change DBVERSION but migrate the data in ACCOUNT every time the DB is opened. (I do not like to always populate data like that...)
 * The second will make the DB changes, PR not to merged yet

davidedelvento:
@gerhardol no need to cancel these two PRs just because I did not follow everything on my first read!

Just to be clear about line 346 (and similarly for the other existing ones). It's not about the compiler. I don't care about the compiler for cases like this. It's not that we are running a triple-nested loop with large counts!
It's all about the reader (and the writer). The simpler the code it is, the easier it is to read, the harder it is to get it wrong, or to introduce subtle bugs which are triggered in edge cases.

davidedelvento:
And yes, I'll be happy to review any PRs you ask me to review (I'll try to review everything, but you are working too fast for me to follow everything on a timely manner, at this moment of my life). 

So let me know if you decide not to cancel these two PRs, I'll give them another look (but as I said, I'll prefer to do so after the whole code is simplified, to avoid having to think about unnecessarily complicated cases).

Thanks!

gerhardol:
@davidedelvento It was not your comments, it was me wanting to merge the changes as much as possible without updating db. (Some of your comments will be included too.)

gerhardol:
Replaced with #611

PR #600, opened 2017-08-05
runalyze login redirect

gerhardol:
No longer possible to login or upload to Runalyze, connecting to runalyze.com version 4.2 (recently updated)
@rickyepoderi The patch is not fully working and disables the account (hides it by default).

Edit: login fails but upload is OK

Based on #593

### DIFF
--- a/app/latest/java/org/runnerup/feedwidget/FeedWidgetService.java
+++ b/app/latest/java/org/runnerup/feedwidget/FeedWidgetService.java
@@ -135,35 +135,6 @@ public class FeedWidgetService extends RemoteViewsService {
                     } else {
                         rv.setViewVisibility(R.id.feed_widget_item_notes, View.GONE);
                     }
-
-                    // c.put(FEED.FEED_TYPE, FEED.FEED_TYPE_ACTIVITY);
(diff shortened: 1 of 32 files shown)


rickyepoderi:
Hi,

I'm on vacation sorry, I just see two things:

* This one was and && on purpose, it always redirects you but when a failure is to login: https://github.com/jonasoreland/runnerup/pull/600/files#diff-a572a7cdd18f7ab95a76dddcf2fa9ae6R321
* When you fail to upload the response is a 200 but with a error message: https://github.com/jonasoreland/runnerup/pull/600/files#diff-a572a7cdd18f7ab95a76dddcf2fa9ae6L525

But I don't know in the new version, I suppose 4.2 is broken without any of these changes, right?

@mipapo Do you know if there's something new in 4.2 to take into consideration?

I'll try to check it but I'm on vacation right now.

Thanks.

mipapo:
@rickyepoderi I'm sorry, but I cannot remember that we changed something in v4.2 which would have that impacts

gerhardol:
@rickyepoderi 
Login seem to work again, I do not know why it did not work for me. (There were redirects, login never completed.)

Issue #334, opened 2015-09-29
Polar HRM (2008, H7) support


tbdal:
Hi Jonas,

thanks for your encredible efforts!

RunnerUp is not able to connect to HRM Monitor by polar from 2008 (called H7?)

Should this HRM work by selecting "Polar WearLink" -> name: Polar iWL ?

Thanks a lot for looking into this, attached you'll find a screenshot of RunnerUp. Can I give you access to logs/debugs in any way?

cheers,
tbdal
![screenshot_2015-09-29-21-10-23](https://cloud.githubusercontent.com/assets/6881189/10175519/9673ef6e-66ef-11e5-910d-fcb1ca70dc7b.jpg)

![Uploading Screenshot_2015-09-29-21-10-23.jpg…]()

tbdal:
i compiled the app. adb shows the following:

09-29 21:41:08.130    6811-6811/? I/art﹕ Late-enabling -Xcheck:jni
09-29 21:41:08.148    6811-6821/? I/art﹕ Debugger is no longer active
09-29 21:41:08.248    6811-6811/? E/org.runnerup.view.MainLayout﹕ app-version: 8000050, upgradeState: SAME, km: true
09-29 21:41:08.252    6811-6811/? W/Resources﹕ Converting to string: TypedValue{t=0x10/d=0x78 a=-1}
09-29 21:41:08.255    6811-6811/? W/Resources﹕ Converting to string: TypedValue{t=0x10/d=0x3e8 a=-1}
09-29 21:41:08.297    6811-6811/? E/org.runnerup.view.MainLayout﹕ Found: /data/data/org.runnerup/files/.., app_workouts, isFile: false
09-29 21:41:08.324    6811-6811/? E/org.runnerup.view.MainLayout﹕ Found: /data/data/org.runnerup/files/../app_workouts, 8-6-4-2.json, isFile: true
09-29 21:41:08.324    6811-6811/? E/org.runnerup.view.MainLayout﹕ Skip: /data/data/org.runnerup/files/../app_workouts/8-6-4-2.json, isDirectory(): false, isFile(): true
09-29 21:41:08.324    6811-6811/? E/org.runnerup.view.MainLayout﹕ Found: /data/data/org.runnerup/files/../app_workouts, MalinEwerlov.json, isFile: true
[...]

Issue #487, opened 2016-11-28
Zephyr randomly disconnects


porzione:
I'm using HR zone target and after random time from 20 to 60 minutes connection is lost. And I'm listening "zero bpm, speed up" until I stop and restart session. This is repeated each time even with full battery.

gerhardol:
FYI: Next version has a message when HR connects/disconnects.

PR #610, opened 2017-08-10
Set WearLink, Zephyr, Samsung4.2 as expermental HRProvider

gerhardol:
There are reports that at least WearLink is not working
All of these should be considered legacy and (unless someone verifies and adds patches for them), be removed in future updates (the BT20 module can be removed completely).
See for instance #334


jonasoreland:
i personally tested/verified them when adding them...

i think it sounds very harsh to remove them just because a few people doesn't get them to work...
with that reasoning we should probably remove all features...

also, iff they don't work, they don't have any negative consequences for something else

gerhardol:
They are not removed, just set as experimental. Existing setups should not be changed
I do not have these devices but the code only uses this in the SettingsActivity

There are several reports since the last years that at least WearLink is not working for users
Zephyr problems in #487 
Samsung4.2BLE should not have many users

Functionality that are known to have problem and no one will fix should at least be hidden.

PR #611, opened 2017-08-11
account independent dbversion

gerhardol:
This is #594, #596 without changes to the database
See previous PRs for comments

To be followed by PR that cleans up the database too
@daviddelvento to review, only last two commits are new

### DIFF
--- a/app/src/org/runnerup/db/DBHelper.java
+++ b/app/src/org/runnerup/db/DBHelper.java
@@ -246,56 +246,51 @@ public class DBHelper extends SQLiteOpenHelper implements
         arg0.execSQL(CREATE_TABLE_FEED);
         arg0.execSQL(CREATE_INDEX_FEED);
 
-        onUpgrade(arg0, 0, DBVERSION);
+        onCreateUpgrade(arg0, 0, DBVERSION);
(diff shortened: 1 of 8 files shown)


davidedelvento:
This looks good to me, simple and clear. 

My only concern regards the two [caught exceptions] and especially the [first one](https://github.com/jonasoreland/runnerup/pull/611/files#diff-7951e0959f06cf40a033bc1a4552ca18R88). 

Are you 100% sure that you might not be catching a wrong exception, mistakenly silencing something that should be more visible, and not risking causing troubles down the road? In the second case, is it ok to return an empty `JSONObject` if the exception happens? I don't have (yet) a full grasp of what these possibilities entail for the whole app, and whether or not it would be better to leave the exception uncaught (or maybe restricted to more specific type, especially the first). Or maybe they are totally fine as is, but I don't know. If you do know they are fine, go for it.

Thanks!

gerhardol:
Updated after the comments, adding printouts. FileSynch configuration will be lost if the exception occurs, better than that the app crashes.

Note: FileSynch config will be lost also when going back from this PR (config is corrupt)

PR #612, opened 2017-08-11
Target Android O

gerhardol:


davidedelvento:
Can't we have different versions of android targeted in travis like it's possible in python? 
A quick look at travis documentation did not clarify...

gerhardol:
.travis.yml must match the build.gradle file
So it is handled in separate branches, one target per branch

gerhardol:
.travis.yml must match the build.gradle file
So it is handled in separate branches, one target per branch

I want to have this in next public build.
The problem with this PR is that the source for O is not released yet so I often go back to sdk25 in the emulator.

gerhardol:
Will close this PR and reopen a new on top of the restructure
A few changes to support Oreo notification channels should be done too.
In addition (not related to target version) is to make sure that a notification is started at background tasks (for instance uploading activities, then switch to another application).

Issue #568, opened 2017-06-15
Zero iteration repeat step causes race condition that crashes phone


sunfish62:
In trying to create a basic interval workout with timed warmup and cooldown, I modified the MalinEwerlov workout in the Advanced tab so that the later interval steps used zero iterations of zero duration. RunnerUp successfully processed the timed warmup and the five iterations of the first interval. 

Once it reached the zeroed out steps, however, my phone completely froze up, and I had to hard boot it. Upon restart, I reopened RunnerUp and opened the workout entry in the History tab. The workout was not accurate in regards to distance and duration (that much I expected). However, opening up the laps page of the workout showed a huge number of zero value laps. I was unable to scroll through all these entries because RunnerUp crashed out again, and my phone reported that RunnerUp was consuming far too many system resources. It is apparent that RunnerUp did not catch the zero iteration setting correctly.

Version 1.57
Android 6.0.1

jonasoreland:
cool!
that should be easy to fix....don't allow zero iterations...

gerhardol:
In pull request, closing

PR #625, opened 2017-08-20
Workout zero repeat crash

gerhardol:
Zero repeat steps will never finish, #568

Still allow 0 repeats, can be used to skip parts 'dynamically'


davidedelvento:
@gerhardol is this the right PR? We need really 141 commits and change 253 files to fix this bug, dropping FROYO as part of the PR, etc? Or was it a typo and you pushed a branch that was supposed to do something else, which should be merged separately?

gerhardol:
This PR is based on #624 that is based on #524 So only the last commit is for this PR, the two before for #624 I cannot test in any other way...

davidedelvento:
Ah, I see. Can't you merge the other two PRs first and do this one later, just to make things easier to follow? 

From a stylistic point of view, why commenting out sections of code instead of deleting them?

gerhardol:
Commenting out:
I do that when there are better ways to do a change. Could hav added comments too

The Asserts are situations that should not occur, but do. Not new.

I could make a new PR sure, but I believe you will see only the relevant commits when the preceeding PRs are merged.
PRs that build on eachother are harder to follow, but is better than big PRs. If the PRs cannot be separate like this that builds on previous, I believe it is fine.

davidedelvento:
Alright, thanks for taking the time to explain.

PR #649, opened 2018-03-23
Material design bottom navigation

12people:
Designed according to the [Material Design bottom navigation spec](https://material.io/guidelines/components/bottom-navigation.html), which is recommended for primary navigation.

Here's what it looks like on Froyo: 
image

It's limited by what TabView allows:

* Uses VectorDrawable only on API 21 and above. If TabActivity was replaced with AppCompatActivity, we could use VectorDrawable on all API versions.
* No animations (like ripples and transitions) and no font size changes. It'd be best to use the official BottomNavigationView element—however, that's not supported on Froyo. We should replace TabView once the minSDK is raised.


12people:
Also, ran optipng on all PNGs in the hdpi folder, which is why they're shown as being replaced. There should be no changes to them other than an optimized size.

gerhardol:
Not completely sure what the guideline recommends compared to top tabs:
https://material.io/guidelines/components/tabs.html#

I cannot find the passage right now, but the double rows of tabs is a no-no and that is still confusing.
I would like to see a standard sidebar instead.
This is an improvement in my opinion never the less. Versions, branch to use and build focus should be solved before a merge though.

@jonasoreland Opinons?

12people:
Difference between the two:
Bottom navigation "provides quick navigation between top-level views of an app." Tabs are more generic, used to " to organize content at a high level, for example, to present different sections of a newspaper".

There are various differences between the two:
- Bottom navigation doesn't support the slide gesture.
- Bottom navigation only allows 3-5 items.
- Bottom navigation is primary navigation only.
- Bottom navigation is mobile-only, translates to a fixed sidebar on the desktop.

As for double tabs being a no-no—if you're talking about standard swipable tabs, then yes, as they make interactions like which tab will be switched when you're swiping more confusing. However, there being a clear distinction between bottom nav and tabs (similar to nav bar vs. tabs), this problem doesn't arise with bottom navigation. In fact, Google uses bottom nav + tabs in some of their apps, e.g. here:
screenshot_20180325-010745

And as a UX designer, I'd **recommend sticking to bottom tabs**. Side navigation has been proven in studies to be less discoverable and provide a worse user experience than visible navigation (whether on the bottom or on top)—e.g. in this [N&N study](https://www.nngroup.com/articles/hamburger-menus/) or [these](https://thenextweb.com/dd/2014/04/08/ux-designers-side-drawer-navigation-costing-half-user-engagement/) [A/B tests](https://medium.com/the-redesign/deconstructing-the-hamburger-menu-9bf4762bde3e).

gerhardol:
Thanks for the motivation.Is this your final proposal for RU UI or just a first step?
Is the UI acceptable after this?
I would probably prefer a floating Add button instead of a Start category

Still, I would like to get more opinions on the UI
@jonasoreland @cmmata @davidedelvento

davidedelvento:
I don't care what it looks like. I was hoping to get more efforts into the functionality instead....

12people:
@gerhardol It's just a first step. It's a deliberate evolution—changing many things at once can be both frustrating for current users and also very hard to get right without conducting rigorous tests.

I'm playing with some mockups here: https://www.figma.com/file/SlPQaT56OztqaLmFZAeM0y/RunnerUp .

I also started work on redesigning the "Configure accounts" page, which will be a bit more disruptive. Expect a PR in a few days.

12people:
@gerhardol I've made sure this works in Froyo, so as master_ng has a minSDK of 21, standard master seems like the fitting place.

(If I was targeting API 21 or later, I would've implemented it differently.)

12people:
Just added two commits related to standardizing theme colors.

I was thinking of doing this as a separate pull request. However, the changes depend on the new bottom navigation bar being approved, so it made more sense to add them here.

The changes concern:
- Setting #003366 as primary color, #53B1FC as accent color (before, no colors were set)
- Running optipng on all PNGs (size reduction with no loss in quality)
- Exchanging android.app.AlertDialog for android.support.v7.app.AlertDialog, so that theme colors are used
- Using a custom base style instead of Theme.AppCompat so that theme colors are used

All changes are quite trivial, see the commits.

gerhardol:
Changes make sense.

gerhardol:
Please rebase before merge

gerhardol:
Thanks!

PR #656, opened 2018-04-24
Android Studio 3.1

gerhardol:
and some related changes


gerhardol:
The build fails due to Lint being more strict, backported the patch from my master_ng branch.

Issue #659, opened 2018-04-29
Current pace totally broken


davidedelvento:
For me the current pace is just a random number in the range half the activity pace to twice the activity pace. Others I've spoke with report similar problem, or even current pace not being reported at all (looking at the source code, it's obvious why and is related to the hardware).

I have a patch which calculates the current pace from last distance and last time, instead of relying on the inaccurate (or unavailable) GPS-based speed. Field tests performed so far show it's much more accurate, and (obviously) available for any device. Would you like a PR?

gerhardol:
If you have a proposal, a pr is always easier to discuss. If it is work in progress, just write that in the PR (for instance wip in the title). 

I do not use current place at all myself but can see problems in the emulator (pause detection is useless partly due to pace).

gerhardol:
I get notifications but I have no computer at the moment - replaced parts but the new was malfunctioning so it is at the shop.

gerhardol:
See #722, #730 for smoothed pace.

gerhardol:
Please try RunnerUp Alpha for the improvements to current pace

PR #660, opened 2018-05-01
Calculating speed by dividing space and time because many Android dev…

davidedelvento:
Calculating speed by dividing space and time because many Android devices do not
implement getSpeed() (even if hasSpeed() is true) or provide incorrect readings. 
See https://github.com/jonasoreland/runnerup/issues/659

### DIFF
--- a/app/src/org/runnerup/tracker/Tracker.java
+++ b/app/src/org/runnerup/tracker/Tracker.java
@@ -83,6 +83,10 @@ public class Tracker extends android.app.Service implements
     public static final int MAX_HR_AGE = 3000; // 3s
 
     private final Handler handler = new Handler();
+    private Double distanceI0 = null;
+    private Double distanceI1 = null;
+    private Long timeI0 = null;
+    private Long timeI1 = null;
 
     TrackerComponentCollection components = new TrackerComponentCollection();
     //Some trackers may select separate sensors depending on sport, handled in onBind()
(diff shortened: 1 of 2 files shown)


davidedelvento:
Just to be clear, I know that this way of calculating "current" speed/pace does not provide the "current" one, but the average since the last time it was called. 

My idea is to push this out and have a few more people try this in various conditions of sky/satellite view (instead of only myself and another person who's currently testing this). The final version should have a hardcoded time in which the "current" value is updated (say 10 seconds or so) which works well with most sky/satellite views, while not being overly long (30 seconds at most, in my opinion). And then of course such a "current" speed/pace is spoken when appropriate with other audio cues.

app/src/org/runnerup/workout/Workout.java
```diff
@@ -523,9 +523,6 @@ public boolean isEnabled(Dimension dim, Scope scope) {
             return tracker.isComponentConnected(TrackerTemperature.NAME);
         } else if (dim == Dimension.PRESSURE) {
             return tracker.isComponentConnected(TrackerPressure.NAME);
-        } else if ((dim == Dimension.SPEED || dim == Dimension.PACE) &&
```
gerhardol: Why remove this?

app/src/org/runnerup/tracker/Tracker.java
```diff
@@ -864,20 +868,24 @@ public Double getCurrentElevation() {
...
-        return getCurrentSpeed(System.currentTimeMillis(), 3000);
-    }
-
-    private Double getCurrentSpeed(long now, long maxAge) {
-        if (mLastLocation == null)
-            return null;
-        if (now > mLastLocation.getTime() + maxAge)
-            return null;
-        double speed = mLastLocation.getSpeed();
-        if ((!mLastLocation.hasSpeed() || speed == 0.0f)
```
gerhardol: Speed from the sensor should be better than "GPS point speed" (I checked a little some years ago). Still, smoothing is required.
RU should offer this as an option.

Extra variables are not needed, mLast2Location has the data already.
Smoothing is required though.

gerhardol:
Super seeded by #716, that makes speed configurable
Worse result for me
Smoothing is required though

PR #664, opened 2018-06-23
Redesign of the history tab

12people:
I redesigned the history tab to be a bit more readable, nicer to look at, and follow Material Design guidelines.

Screenshot from Froyo:
image


gerhardol:
Looks good

Marx2:
But fits only a few activities. Maybe smaller icons?

niedz., 24.06.2018, 02:00 użytkownik Gerhard Olsson <
notifications@github.com> napisał:

12people:
@Marx2 
Tell me a bit about the way you use the history overview, so that I can adjust the design to suit your needs as well. When do you find yourself looking at the history overview and for what reasons?

gerhardol:
When looking at the history, it is nice to get an overview, not just the latest day.
The icon uses more space than the information right too.

I do miss the time of day too.
But in general it looks better

12people:
I personally use the history overview to find a run on a specific date as well as to get an overview of how many days pass between my runs (I'm not always consistent). The time of day isn't important for me at all. That's why I'm sorting things by day.

However, I realize your use cases may be different, so I'd like to discuss them. When you open the history tab, what are you generally trying to do? What information are you looking for? How many entries do you have in the tab? How many per day? How much does the type of sport vary for you?

12people:
Now I'm thinking—for future design changes, it'd be useful for me to have a good idea of your use cases. In the UX design process, one generally conducts user interviews and/or observations and then makes personas based on those.

Would any of you guys be willing to be interviewed? Ideally in a public video that we'd post on YouTube for future reference?

gerhardol:
You can interview me, but I am probably not the average user - I do not use RU much right now as my phone will not handle +3h any more with GPS.
My primary use of RU has been as a data logger (getting all available information). I had some intention to get a smartwatch to use with the phone, but the options I expected to be available has even dried out the last years. Instead Garmin has both got more dominant and more expensive. But I want to have an open source SW so I have options.

Hoping for more phones with GPS L5 soon (just Xiaomi Mi8 so far).

So summary of use:
 * Data logger: Start and put away when using, no feedback during workout. In unknown areas, I could compare to the map too (not necessarily done in RU.)
 * After exercise: Upload to service or file, view details there. Some basic checks of the exercise afterwards on the phone, maybe upload to optional services.

Marx2:
Usually I look at last training with more details, but I'm also looking at
overall progress. I would love to have custom view, user selectable
columns, because I'm sure everybody use it differently. For example I would
like to see medium heat rate, but somebody who cycle, can prefer cadence. I
prefer table view with small font and many values, someone can prefer icons
(while pretty, they give no value for me because I only run). We can also
make different view, of rotated (view is wide so can fit more info)

niedz., 24.06.2018, 13:07 użytkownik Gerhard Olsson <
notifications@github.com> napisał:

12people:
@Marx2 I'd like to come up with a design that works for most people right from the start. Customization would be there only to cover any non-standard niche cases.

If I assume that it's important for most to:
- get to the last training quickly
- compare distance, time, and speed between consecutive days
- see an overview of all types of exercises

... and for runners to see and compare HR between consecutive days, while for cyclists it's to see and compare cadence between consecutive days, I might get to a default design like this:

history

The only question is, are these assumptions correct? Are there any important ones missing? User interviews (and ideally observations, too) would be useful for that.

gerhardol:
I like the mockup slightly better, but the left icon is consistent with most designs. Left icon makes it easy to select multiple activities (no use for that right now).

A smaller left icon and info on the second line is maybe a better setup?

Marx2:
New design is quite nice. What if there's no heart rate available?

niedz., 24.06.2018, 14:45 użytkownik Gerhard Olsson <
notifications@github.com> napisał:

12people:
@Marx2 It just wouldn't be shown

12people:
Should be working, new layout should be final

This is what it looks like in Froyo:
image

PR #665, opened 2018-06-24
More readable advanced steps

12people:
A more readable presentation for workouts. The goal is to be able to read the workout at a glance.

Here are screenshots from Froyo:

Advanced tab:
image

Workout editor:
image


12people:
New screenshots from Marshmallow, after some changes:

screenshot_20180624-124322
screenshot_20180624-124605

gerhardol:
With this change, you will have to know what a symbol means, that Play symbol is active for instance. 
Is this clear enough for every one?
(It looks much better though).

12people:
@gerhardol That's why the latest commit. When you don't have a concrete goal, the label is shown, so you know approx. how fast to run. When you do have a concrete goal, you know exactly at what speed to run, so the type is redundant and self-explanatory.

12people:
That said, it'd be ideal to test this by giving a runner a few tasks:
1) Start a run with intervals of decreasing length of 8, 6, 4, and 2 mins and slightly increasing speed, with rests of 2 mins in between.
2) Create a workout with two intervals—8 mins and 16 mins, with heart rate in zone 5 during. There should be a warm up at the start, a cool down at the end, and a pause in between.

When testing, let the person accomplish these tasks without any interference or help. Ask them to voice every thought out loud (and remind them if they forget to). That should tell you when a person is confused and why.

For the purpose of a valid test, don't draw unnecessary attention to the icons or ask about them during the test. After the test, you may ask if they found anything confusing and, potentially, as a follow-up question, what are their thoughs on the icons (without implying the icons are confusing).

Anyone willing to try this out is welcome to. :)

12people:
@gerhardol Had an idea of how to make the icons clearer:
image
(It's only a mockup, not implemented yet.)

What do you think?

12people:
Implemented based on the newer, clearer design.

Here's what it looks like in Froyo:
image

12people:
@gerhardol Alright to merge?

app/src/org/runnerup/view/StepButton.java
```diff
@@ -100,33 +94,61 @@ public void setOnChangedListener(Runnable runnable) {
...
-            mGoalRow.setVisibility(View.GONE);
+            mGoalValue.setText(step.getIntensity().getTextId());
         } else {
-            mGoalRow.setVisibility(View.VISIBLE);
-            mGoalType.setText(res.getString(goalType.getTextId()));
-            mGoalValue.setText(formatter.format(Formatter.Format.TXT_SHORT, goalType,
-                    step.getTargetValue().minValue)
+            String prefix;
+            if (goalType == Dimension.HR || goalType == Dimension.HRZ)
+                prefix = "HR "; //todo should use a string
```
gerhardol: If there is no usable string, add one

gerhardol:
Can you fix conflicts?

PR #669, opened 2018-06-29
Material app bar for history items

12people:
Another iterative change to RunnerUp's look. Here's what it looks like in Froyo:

image

The long-term goal is to get something like this:
image

However, as listening for sport changes may be tricky and as swipeable tabs are a big task (that would call for using a ViewPager and putting individual tab contents into fragments), I'm leaving those for pull requests sometime down the line.

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
   ~ Copyright (C) 2012 - 2013 jonas.oreland@gmail.com
   ~
   ~  This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


12people:
Also, as a slight tweak, the Save button is now hidden unless editing.

app/AndroidManifest.xml
```diff
@@ -17,145 +16,173 @@
...
+        <activity
+            android:name=".view.AudioCueSettingsActivity"
+            android:theme="@style/AppTheme" />
+
+        <activity
+            android:name=".view.DetailActivity"
+            android:configChanges="orientation|screenSize|uiMode"
+            android:parentActivityName=".view.MainLayout"
+            android:theme="@style/MainLayoutTheme">
+            <!-- Parent activity meta-data to support 4.0 and lower -->
```
gerhardol: Is this the only changed info?

app/res/values/styles.xml
```diff
@@ -88,8 +88,14 @@
...
 
+    <style name="AppBarStatisticsText">
+        <item name="android:fontFamily" tools:ignore="NewApi">@string/fontFamily_medium</item>
+        <item name="android:textSize">16sp</item>
+        <item name="android:textColor">@android:color/white</item>
+    </style>
+
     <style name="ListCategoryHeading">
-        <!-- item name="android:fontFamily">@string/fontFamily_medium</item -->
+        <item name="android:fontFamily" tools:ignore="NewApi">@string/fontFamily_medium</item>
```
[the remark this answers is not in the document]
12people: It works. It's really just about using the medium font weight rather than regular on devices that support it. Not a significant difference.

app/AndroidManifest.xml
12people: Along with these lines that follow, yes:
+                <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value=".view.MainLayout" />
+        </activity>

Sorry for the auto-indent—I'm used to doing auto-indent with every change, this may be the first time when git diff marked almost the entire file.

gerhardol: It is easier to review if refactoring are separate commits

gerhardol:
Can you fix the conflicts?

12people:
@gerhardol Done

BTW, not sure where to ask, so I'll ask here:
I'm having problems with
1) Android Studio not seeing the v4 support and Design libraries. The build works fine, but it's harder to read through code because the IDE is underlining things as errors that it shouldn't be.
2) not being able to use the Floating Action Button. I keep getting the message "You need to use a Theme.AppCompat theme (or descendant) with the design library." I'm trying to add it to the History tab, I've change HistoryActivity to extend AppCompatActivity, added the theme that inherits from AppCompat for the activity in AndroidManifest, but still no luck.

gerhardol:
@12people 
Thanks for the update

For Froyo, such problems are one of the reasons I don't like to support it, it adds to much time. Also, I have no device to test it on (emulator is not really an option with AMD cpu, I missed the haxm difference when replacing the PC after an emergency breakdown).
To be clear, as long as someone support Froyo, I have nothing against supporting it officially.

Generally, I develop for 'latest' then adjust compatibility errors and briefly test in Froyo emulator.

For the problems:

No specific lead 
I assume you set build variants.

Froyo uses support lib 24.1.0 (latest with Froyo support) while 'latest' uses 27.1.1 (matching the target/compile SDK version). Maybe Froyo needs to adjust the target/compile SDK too?

It may be  that AS3.1 somehow is limiting the support for Froyo too.
I would like to use jetpack/androidx and app bundle in AS3.2 when it is released. If you (or someone else as active as you) want to continue develop for Froyo in the master branch, that will be delayed. But it would be interesting to hear from you if it is working.

PR #673, opened 2018-07-01
Adjust default values to more sensible values

### LINKED ISSUE #621: Default values review
There are a number of default values in RU that possibly should be changed.
The settings creates question from users, it may be better to disable some functionality by the default and let users enable them from the myriad of settings (in my opinion a good thing).
(There are hardcoded settings that could be made visible and changed too.)

I listed a few issues with discussions to these, there are more

Auto pause (currently On, suggest Off)
The setting should not ba a problem but the pace smoothing is not optimal and will occasionally give a lot of pauses.
#598 

GPS sampling (Currently 0.5s, 0m; suggest 2s, 10m)
GPS is unreliable and will measure longer distances the worse the GPS is. For me, I get 5-15% longer than my running watches (a trusted Garmin Forerunner and a Suunto Ambit 3). Newer Garmins have mediocre GPS but often make a good job in the software to measure distances. A good filter that uses accelerometer etc is very hard, but using minimum 2s and 10m gives decent result in my limited tests. Other apps that seem to use such filtering is Runkeeper and Ghostracer that both are more consistent than RU default and for instance Strava.
#598 #464 #257 

Auto start GPS (currently On, suggest off)
To explicitly start the GPS gives more intuitive functionality than to keep GPS on when the app is in the foreground.
#255 

Step sensor for cadence (Currently off, suggest On)
Was not enabled when I added support, probably a good thing
(Keep Temperature and Barometer off?)

Enable Live
Hide the setting (unless avanced is ticked?), no working server?
#351

gerhardol:
fixes #621

### DIFF
--- a/app/res/xml/settings.xml
+++ b/app/res/xml/settings.xml
@@ -32,7 +32,7 @@
         android:key="sensors_preferencescreen">
 
         <CheckBoxPreference
-            android:defaultValue="true"
+            android:defaultValue="false"
             android:key="@string/pref_startgps"
             android:persistent="true"
             android:title="@string/Auto_start_GPS" />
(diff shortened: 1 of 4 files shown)


gerhardol:
@12people 
I will appreciate if you look at the PRs

gerhardol:
@12people 
The projects on GitHub is not what I do for living, the SW I work with is completely diffrent. So I am an amateur here too.
Feedback is good

Issue #640, opened 2018-02-11
Save activity button is remains disabled after entering all details of workout.


iamank1t:
#### Expected behavior
After entering details about my workout(like duration, distance and pace), the save activity button needs to be in enabled state so that i can save my activity.

#### Actual behavior
I tried to save  workout in app. After filling all details(sport, duration, distance, pace) the save activity button remains in enabled state for 1 or 2 second and then again become disabled. 

#### How to reproduce
Download RunnerUp application from Android playstore. Open app, go to manual tab(last tab in your right side) under Start tab. Now do the following steps -
- Choose any "Sport".
- Don't choose any "Date", keep it as it is.
- Don't choose "Start Hour", keep it as it is.
- Set a "Duration".
- Set a "Distance".
- Set a "Pace".
- Add workout note if you want or keep it as it is.
Now after following this steps you will see the button remain green(enabled) for some second and become inactive again after 1 or 2 sec. 
 <pre>
<code>
* Device:  Samsung J7 Prime
* Operating system: Android 7.0 Nougatt
* App Version: v1.58
</code>
</pre>

#### Recording Of The Bug
https://www.youtube.com/watch?v=AIR4q4gnFrU
    
While playing video keep eye on the save activity button at bottom.

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@iamankit/save-activity-button-is-remains-disabled-after-entering-all-details-of-workout">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

gerhardol:
Closed with #677

PR #677, opened 2018-07-04
Move manual entry to history tab

12people:
I've taken the liberty of moving manual exercise entry to the history tab. I'm proposing this for 2 main reasons:
1. It makes more sense conceptually. One is adding a manual entry to history, not starting a manual run. Therefore, placement in the history tab makes more sense.
1. It paves way for a nicer UI on the Start tab. I made some mockups [here](https://www.figma.com/file/SlPQaT56OztqaLmFZAeM0y/RunnerUp).

As a side effect, it makes code more legible and maintainable—there were lots of hacks to make manual entry work with the rest of the Start entries.

screenshot_20180704-230054
screenshot_20180704-230100

It uses a FAB look-alike on Froyo and a true FAB on API 15 and above. (I somehow couldn't get design library widgets to work with API 15 and below, kept getting an error saying that the theme does not inherit from Theme.AppCompat, though it clearly does.)

Here's a Froyo screenshot:

image


gerhardol:
I would like to have + button available everywhere. Normal press to start activity, long press to get other alternatives.
(I am fine with this change though, I do not want to slow you down.)

12people:
I agree with having a FAB in more places—that's the next step! :)
Please no long press, though—it's not discoverable.

gerhardol:
Looks OK. As this is a change that looks almost the same, the adding of manual activities is a little hidden. Any thoughts?

Some new lints though. Fix before merge (this time).
(You can look in Travis or run lintLocalRelease)

12people:
For current RunnerUp users, it will be a change, but I hope they'll find manual entry quickly. (There aren't many screens to go through.) On the plus side, for users coming from Runtastic or Endomondo, it'll be where they're used to it.

Either way, this is an important change for me. If you look at my mockups, you'll see what the end goal is. It wouldn't be possible without separating manual entry.

Lints should be fixed. Will remember to run it next time.

12people:
Also, this fixes bug https://github.com/jonasoreland/runnerup/issues/640

gerhardol:
Thanks
(One lint remains, someone also need to update lint for wear, so it is easy to see that there are no remaining issues. Most lints are actually not interesting, but some are really useful.)

PR #679, opened 2018-07-05
Feed tab redesign

12people:
A redesign of the feed tab:

screenshot_20180705-153018

An effect of the changes proposed is an action bar on the Start and History tabs. I've been meaning to implement such a thing anyway, but it became a necessity with this PR for the placement of new "Edit accounts" and "Refresh" buttons.


gerhardol:
Quick check: Looks good
There is a lint for overdraw.

A good motivation for the last three entries in the activity menu?
 Settings, News and Review RU
Settings would be important if the bottom tab menu is removed (or Settings removed from the existing, which I prefer to not do). News/Review is thrown to the users at first startup, should be enough.

12people:
I'll check the lint out and resolve conflicting files.

The activity menu has actually always been there—it just became visible with the introduction of the action bar. I'd like to get rid of the menu altogether. Can I do it in this PR?

gerhardol:
> I'd like to get rid of the menu altogether. Can I do it in this PR?

The menu were added for pre 4.0 devices with menu button.
The first three are useful (but the History tab would have separate items), the configuration is maybe not obvious?

I will leave the decision to you!

12people:
@gerhardol I solved it the way I originally wanted to solve it: by putting the buttons in the most fitting places.

"Manage audio cues..." is now under the "Audio cue settings" dropdown, "Manage workouts..." is now under the "Wourkout" dropdown.

app/build.gradle
```diff
@@ -122,6 +122,8 @@ dependencies {
     //noinspection GradleDependency
     froyoImplementation 'com.android.support:design:24.1.0' //SDK <10 dropped in 24.2
 
+    latestImplementation 'com.android.support:cardview-v7:27.1.1'
```
gerhardol: latestImplementation "com.android.support:cardview-v7:${rootProject.ext.supportLibrary}"

common/src/main/res/values/strings.xml
```diff
@@ -134,6 +134,8 @@
   <string name="Manage_accounts">Manage accounts</string>
   <string name="Manage_audio_cues">Manage audio cues</string>
   <string name="Manage_workouts">Manage workouts</string>
+  <string name="Manage_audio_cues___">Manage audio cues…</string>
```
gerhardol: Why add new strings instead of reusing the existing, adding elipsis?

PR #681, opened 2018-07-05
Lint update for Froyo and wear

gerhardol:
@12people 
Suppressed widgetCategory from recent update

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ notifications:
   email: false
 
 script:
-    - ./gradlew lintLatestRelease assembleLatestRelease wear:lR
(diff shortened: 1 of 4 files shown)


gerhardol:
Some proguard rules should be updated too

12people:
Aside from the change in feed_widget_provider_info.xml, all the other changes seem to be computer-generated to me.

All looks good to me, though I'm only skimming through the computer-generated parts.

gerhardol:
The baseline file is updated by removing it and running lint again
However, you have to make the decision if the suppressing is correct in the baseline file

Will not handle proguard in this PR

PR #691, opened 2018-07-23
Autolap: Trigger at most likely point rather than always after the limit

gerhardol:
The autolaps are currently drifting over time, this will keep the average closer to the configuration

Note: Similar could be done for workouts


gerhardol:
An alternative/addition could be to set the next expected limit. 
I.e. if the current autolap was at 1011 m, next expected limit would be 989 m after the previous.

PR #693, opened 2018-07-29
Start screen relayout

12people:
A new layout for the start screen:
screenshot_20180729-184137
screenshot_20180729-184210

To create the sport spinner in the ActionBar, I needed to abstract the TitileSpinner code to be able to make spinners that have the same behavior, but a different look. The result is splitting up the TitleSpinner code into a view and a presenter, with an interface that defines the communication from presenter to view.

I'm deliberately not changing the look of the tabs—in the future, it'd be nice to use native, scrollable tabs instead. That would require splitting tabs into fragments, which is a rather big undertaking.

After this PR is accepted, I'm in favor of doing a new alpha release as well as raising the minSDK to ICS for later releases.


app/res/layout/start.xml
```diff
@@ -15,144 +14,121 @@
...
+        </LinearLayout>
+
+    </android.support.v7.widget.Toolbar>
+
     <TabHost
         android:id="@+id/tabhost_start"
         android:layout_width="match_parent"
-        android:layout_height="match_parent"
+        android:layout_height="0dp"
+        android:layout_above="@+id/gpsinfo"
```
gerhardol: Lint on this and next

F:\dev\gc\runnerup\app\res\layout\start.xml:79: Warning: Invalid layout param in a LinearLayout: layout_above [ObsoleteLayoutParam]
        android:layout_above="@+id/gpsinfo"
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
F:\dev\gc\runnerup\app\res\layout\start.xml:80: Warning: Invalid layout param in a LinearLayout: layout_alignParentTop [ObsoleteLayoutParam]
        android:layout_alignParentTop="true"

app/res/layout/start.xml
```diff
@@ -15,144 +14,121 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="right"
+            android:layout_marginRight="16dp"
+            android:background="@drawable/indicator_bg"
+            android:gravity="center_vertical"
+            android:orientation="horizontal"
+            android:padding="8dp">
+
+            <ImageView
```
gerhardol: Suppress lint


F:\dev\gc\runnerup\app\res\layout\start.xml:53: Warning: Missing contentDescription attribute on image [ContentDescription]
            <ImageView
            ^
F:\dev\gc\runnerup\app\res\layout\start.xml:60: Warning: Missing contentDescription attribute on image [ContentDescription]
            <ImageView
            ^

app/src/org/runnerup/view/StartActivity.java
```diff
@@ -96,49 +99,50 @@
...
-
-    private Formatter formatter = null;
+    private View startButton = null;
+    private ViewGroup gpsLayout = null;
+    private TextView gpsMessage = null;
+    private Button gpsEnable = null;
+
+    private TextView gpsIndicator = null;
+    private View hrIndicator = null;
+    private View watchIndicator = null;
```
gerhardol: These are WearOS related, could be something else than a watch and all watches are not WearOS.

Have you tested with a WearOS device?
(Can be emulated).

app/src/org/runnerup/widget/SpinnerPresenter.java
```diff
@@ -0,0 +1,605 @@
+package org.runnerup.widget;
```
gerhardol: Please add a copyright header (missing for some existing files but new files should have them).

app/src/org/runnerup/widget/TitleSpinner.java
```diff
@@ -17,665 +17,158 @@
...
-         * @return
-         * @throws java.lang.IllegalArgumentException
-         */
-        int preSetValue(int newValue) throws java.lang.IllegalArgumentException;
-    }
-
-    public interface OnCloseDialogListener {
-        void onClose(TitleSpinner spinner, boolean ok);
-    }
+@TargetApi(Build.VERSION_CODES.FROYO)
```
gerhardol: Remove this line (done in all other files in the Lint cleanup, you based it on older files)

app/src/org/runnerup/view/StartActivity.java
```diff
@@ -96,49 +99,50 @@
...
+    SQLiteDatabase mDB = null;
+
+    Formatter formatter = null;
     private NotificationStateManager notificationStateManager;
     private GpsSearchingState gpsSearchingState;
     private GpsBoundState gpsBoundState;
     private boolean headsetRegistered = false;
 
-    /** Called when the activity is first created. */
+    /**
```
gerhardol: Java doc tag, change to single line comment //

app/src/org/runnerup/view/StartActivity.java
```diff
@@ -344,14 +367,14 @@ public void onPause() {
         super.onPause();
 
         if (getAutoStartGps()) {
-            /*
+            /**
```
gerhardol: Java doc tag, change to single line comment //

app/src/org/runnerup/view/StartActivity.java
```diff
@@ -581,19 +598,15 @@ private Workout prepareWorkout() {
...
-            if (!mGpsStatus.isEnabled()) {
-                startActivity(new Intent(Settings.ACTION_LOCATION_SOURCE_SETTINGS));
-            } else if (mTracker.getState() != TrackerState.CONNECTED) {
-                startGps();
-            } else if (mTracker.getState() == TrackerState.CONNECTED) {
+            if (mTracker.getState() == TrackerState.CONNECTED) {
                 mGpsStatus.stop(StartActivity.this);
 
-                /*
+                /**
```
gerhardol: Java doc tag, change to single line comment //

app/src/org/runnerup/view/StartActivity.java
```diff
@@ -581,19 +598,15 @@ private Workout prepareWorkout() {
...
                  * unregister receivers
                  */
                 unregisterStartEventListener();
 
-                /*
+                /**
```
gerhardol: Java doc tag, change to single line comment //

app/res/layout/start.xml
```diff
@@ -15,144 +14,121 @@
...
+
+    <android.support.v7.widget.Toolbar
+        android:id="@+id/start_actionbar"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:background="@color/colorPrimary"
+        android:elevation="4dp"
+        tools:ignore="UnusedAttribute">
+
+        <org.runnerup.widget.ClassicSpinner
```
gerhardol: This is the only instance where ClassicSpinner is used. Can it be replaced?

12people:
@gerhardol 
To reply to various comments:

> I am not fond of hiding the GPS info when connecting. The accuracy (can still see sats) is something I use before taking off to make sure data will be good. Similar for elevation, stabilizing the barometer elevation.

As a layperson in this area, while I understand what the numbers mean, I have no idea what is a good accuracy, what impact elevation has, how to tell when the barometer is stable, or how many GPS satellites is a good amount. I, personally, just need to know whether GPS is precise, acceptable, inaccurate but usable or not active at all.

Would these categories be too broad for you? If I'm understanding you correctly, you need to have even more precise knowledge about the GPS because they let you make better decisions.

If so, it seems like we have two different personas here. I'm assuming that the less knowledgeable persona is more common, which is why the detailed info was relegated behind a toast. I've looked at other running apps and they generally don't provide detailed GPS info, which suggests that the second persona is indeed more common.

To satisfy both, I propose an expandable status section at the bottom like this:

basic
basic-1

I'm guessing satellite count isn't the best way to gauge GPS status. Could you give me a primer on what constitutes a) precise, b) acceptable, and c) inaccurate but usable levels of GPS, so that I can implement the 3 bar system as shown in the mockup?

> I want to see what HRM sensor that the phone tries connect to so it does not grab a sensor I do not want this session.

This would only be useful for people with multiple HRM sensors and only in the instance that they have several of them connected to the phone at the same time. Also, to be confused about which device RunnerUp is reading, they would have to have set up several in RunnerUp already.

To me, this seems like a very rare situation, if at all realistic. Has anyone run into this before?

Either way, the expanding section I proposed should solve this.

> To repeatedly tap and see a toast is not my preference here. It could be OK if the toast appeared periodically (a new preference or tap once to see messages, second time turning off) if you have arguments to not show the status by default. (The floating button uses less screen space and looks better)
> Edit2: I will not block merging due to this.

See above.

> 
> The play button is very inconsistent with the other activity buttons. Change all at the same time?

Which activity buttons do you mean?
RunnerUp has always used the play icon for starting or resuming a run, pause for pausing it, stop for ending it.

The "Start" tab in the bottom navigation is different--I consider that more akin to the Start menu in Windows than a "Start activity" button. IMHO, it shouldn't have the same icon because its functionality is rather different. Perhaps we should use the official RunnerUp logo for the bottom tab?

> Edit: The color in the toolbar is the standard RU dark blue and looks a little dated. I would like to change that to something else. Maybe gray, maybe lighter blue. The app icon should be replaced too...

I'd like to keep the RunnerUp blue, but use specific activity colors for activity-related screens. For example, for run-related screens, use green. For biking-related screens, orange. You can see that reflected in my mockups.

If you'd like to play around with the design yourself, you're welcome to duplicate my Figma project: https://www.figma.com/file/SlPQaT56OztqaLmFZAeM0y/RunnerUp?node-id=575%3A649 .

> This is the only instance where ClassicSpinner is used. Can it be replaced?

The reason for refactoring TitleSpinner was so that I can have a spinner that looks like a title. Yes, this is the only place it's used, but it's quite important to the design. Its code is quite lean, so it shouldn't pose much of a maintenance burden.

12people:
As for the code changes, I'll get to them when I redesign the status indicators.

Let me know if the suggested expandable status bar would work and how to go about the simplified GPS indicator.

gerhardol:
For advanced vs normal user: 
Most users may not care, but I am probably advanced in that way that I prefer to make to decisions myself. The details is a reason why I want to keep RU, then the interfaces are not simplified when not needed. I agree that the information does not have to be "in the face" for a "normal" user.
As for the sensor, my case is is currently a special one, but if more types of sensors (cadence, cycling etc) the design should somehow allow for this.

The Additional status row looks good. The mockups with cues in the toolbar seem best to me. Toggle between normal/advanced status with clicks?

For "normal user" accuracy I propose the following:
 * red No GPS
 * red Bad GPS accuracy > 15m
 * orange Acceptable 7-15 m
 * green Good GPS <= 7 m

Use a small signal strength bar instead of colors?

I am not completely sure that the Acceptable category  should be included (so Good < 10m, bad > 10m) or that the limits are the best. I would not start an activity with less than 7m accuracy, but even 10m can be difficult in city canyons. 

Note: Satellite count is the only way to get information before a fix. For instance, I have bad GPS inside my apartment and have to stretch to the window to see satellites at all bad days.

> Perhaps we should use the official RunnerUp logo for the bottom tab?

Can do.
I like the mockups, for instance removing bottom text. I would prefer to update the RU color before using it in the bottom navigation bar, it is a little "murky" right now. The color and icon looks very dated when comparing to other apps on newer phones that use round icons by default. (I believe round icons wastes space, but mixing round/square is ugly.) Separate issue, but a new icon for 2.0 would be nice.

>> The play button is very inconsistent with the other activity buttons. Change all at the same time?

> Which activity buttons do you mean?

The start button is now a round "play" button, standard in Android. The ongoing activity buttons are square. I backpedal on the suggestion to change all at the same time though.

Most of the discussions here are further enhancements and should be done in a follow up PR.

gerhardol:
> If you'd like to play around with the design yourself, you're welcome to duplicate my Figma project: 

I am bad at making mockups myself (mostly command line person), but I have looked at the mockups.
I saw you had a question about laps vs intervals.
Laps are started with lap button or autolap and intervals are workout triggered.

In most apps/watches, they are the same. In Suunto autolaps are separate but manual laps and "intervals" are the same.

gerhardol:
Release build (alpha) planned for this weekend, this PR should be included (cleanups, I assume that the refresh will not be ready).

Some proguard warnings related to the changes
I am not sure if the messages can be suppressed or if the design should be changed.


Note: the configuration keeps the entry point 'org.runnerup.view.StepButton { void setStep(org.runnerup.workout.Step); }', but not the descriptor class 'org.runnerup.workout.Step'
Note: the configuration keeps the entry point 'org.runnerup.widget.ClassicSpinner { void setViewAdapter(org.runnerup.widget.DisabledEntriesAdapter); }', but not the descriptor class 'org.runnerup.widget.DisabledEntriesAdapter'
Note: the configuration keeps the entry point 'org.runnerup.widget.TitleSpinner { void setViewAdapter(org.runnerup.widget.DisabledEntriesAdapter); }', but not the descriptor class 'org.runnerup.widget.DisabledEntriesAdapter'
Note: the configuration keeps the entry point 'org.runnerup.widget.TitleSpinner { void setOnSetValueListener(org.runnerup.widget.SpinnerInterface$OnSetValueListener); }', but not the descriptor class 'org.runnerup.widget.SpinnerInterface$OnSetValueListener'
Note: the configuration keeps the entry point 'org.runnerup.widget.TitleSpinner { void setOnCloseDialogListener(org.runnerup.widget.SpinnerInterface$OnCloseDialogListener); }', but not the descriptor class 'org.runnerup.widget.SpinnerInterface$OnCloseDialogListener'

app/res/drawable/ic_heartrate_white_24dp.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
12people: You can include them in an ImageView somewhere. Not sure if there's a simple way to view android vector drawables.

app/src/org/runnerup/widget/SpinnerPresenter.java
12people: Added

12people:
In the end, I didn't use the RunnerUp icon for Start--I felt like it didn't really fit.

gerhardol:
I have a few comments on the implementation (there are new lints too) but will merge this as is and take changes separately.

Also have some ideas for how the start activity should look like Upper tabs with Standard/Interval/Advanced can then be removed

Nice work, thanks!

PR #699, opened 2018-07-31
Workouts: Share/Import did not work

gerhardol:
Share and import failed due to the handling of .json suffix, missing in some situations
Space missing in some printouts

Note:1 Import still "hangs" in some situations, probably Android rather than RU
Note2: RU must be stopped (swiped away) for the import to work
Note3: Manage Workouts-My Phone should be expanded by default. @12people ?


gerhardol:
> What are the other possible entries on the page other than "my phone"?

Nothing now. Used to be Garmin download.

12people:
Alright. In the future, it'd be good to get rid of the section entirely and just show the list items.

PR #710, opened 2018-08-05
Version code and changes.html update for Alpha release

gerhardol:
Version code and changes description for 2.0 alpha

### DIFF
--- a/app/assets/changes.html
+++ b/app/assets/changes.html
@@ -5,12 +5,48 @@
 </head>
 <body>
 <h1>What's new</h1>
+<h2>v1.90.1</h2>
+<p>First alpha release for 2.0. This build appears as a separate application and should be possible to use in parallel to the standard Play build.
+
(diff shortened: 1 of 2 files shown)


gerhardol:
@12people I left the description for the GUI update short. Any suggestions?
@jonasoreland Any comments? (Still awaiting #693 before an alpha).

12people:
@gerhardol Short is fine. If you wanted to, you could mention that only parts have been updated. There are other parts, like the activity UI or the workout manager, that haven't been updated at all.

Glad to hear F-Droid support is returning!

Surprised to hear about 5.0 being the minSDK in the future. This would cut out approx. 15% of Android users. I personally would prefer raising the minSDK to API 15, which would only cut out around 0.3% of people. Strictly design-wise, it's no burden at all--even Froyo support was surprisingly manageable, despite it being obsolete.

gerhardol:
changes.html will be updated before the release

PR #711, opened 2018-08-05
Refreshed product icon

12people:
Rather than create an entirely new icon (which would take a lot of time and which I don't feel like doing right now), I'd simply like to refresh the current icon and use adaptive icons on newer devices.

Here's an icon for Google Play:
ic_launcher-web


12people:
I see that I unwittingly based the changes on previous work.
Is there a simple way to rebase that last commit?

PR #713, opened 2018-08-05
Start screen review

gerhardol:
Review comments for #693 (and related)

### DIFF
--- a/app/proguard.txt
+++ b/app/proguard.txt
@@ -1,19 +1,32 @@
+#Suppress warnings for proguard, for all irrelevant warnings
+
 -dontwarn com.dsi.ant.**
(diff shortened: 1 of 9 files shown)


gerhardol:
@12people 
Still two more Lints that should be fixed

One is just a supress in Froyo
The second I am not completely sure of tonight
https://stackoverflow.com/questions/8318765/how-do-i-use-a-compound-drawable-instead-of-a-linearlayout-that-contains-an-imag

12people:
I guess you fixed it! Thanks :)
I understand the compound drawable issue, but since there's various visibility changes in the code, it would take a while for me to reimplement using those. I think ignoring this lint check should be alright for now.

PR #716, opened 2018-08-09
Current pace and autopause

gerhardol:
Various tweaks for current pace

* Pace and HR is now displayed during a run
* Code mixed GPS and system time, if system time had drifted, current pace/hr could be considered invalid. Autopause could both trigger more often or not at all.
* Autopause refactoring, current code was hard to understand


Uatschitchun:
If I get it right, this PR is where calculate speed from GPS Points was introduced, correct? What is the difference, when activating this in the settings? How is speed calculated otherwise?
I would be willing to add an explanation to the wiki then!

gerhardol:
The speed is normally calculated from last to current GPS point (so distance last location divided with time). The GPS may report speed with better accuracy as a direct number.
I have not seen much difference myself.

PR #720, opened 2018-08-11
Small string tweaks for better clarity and consistency

12people:
While translating, I ran into strings that weren't very clear, that were using inconsistent capitalization, weren't gramatically correct, etc.

This is a proposal for a fix.


common/src/main/res/values/strings.xml
```diff
@@ -169,44 +169,44 @@
...
   <string name="Advanced_options">Advanced options</string>
   <string name="Mapbox_default_style">Mapbox style</string>
-  <string name="RunnerUp_live_address">RunnerUp live address</string>
+  <string name="RunnerUp_live_address">RunnerUp Live address</string>
   <string name="Autopause_after_s">Autopause after (s)</string>
   <string name="Autopause_min_pace_minkm">Autopause min pace (min/km)</string>
   <string name="Convert_rest_on_Interval_tab">Convert rest on Interval tab</string>
   <string name="Convert_rest_on_Advanced_tab">Convert rest on Advanced tab</string>
-  <string name="Target_pace_moving_average_seconds">Target pace moving average seconds</string>
+  <string name="Target_pace_moving_average_seconds">Target pace calculation frequency (s)</string>
```
gerhardol: I cannot see this as an improvement

common/src/main/res/values/strings.xml
```diff
@@ -169,44 +169,44 @@
...
   <string name="Use_your_headset_to_startpauseresume_RunnerUp">Use your headset to start/pause/resume RunnerUp</string>
-  <string name="Convert_reststep_with_type_distance_to_recoverystep_for_Intervaltab">Convert rest-step with type distance to recovery-step for Interval-tab</string>
-  <string name="Convert_reststep_with_type_distance_to_recoverystep_for_Advancedtab">Convert rest-step with type distance to recovery-step for Advanced-tab</string>
-  <string name="Export_database_to_sdcard_eg_for_upgrade">Export database to SD-card (e.g for upgrade)</string>
-  <string name="Import_database_from_sdcard_eg_after_upgrade">Import database from SD-card (e.g after upgrade)</string>
+  <string name="Convert_reststep_with_type_distance_to_recoverystep_for_Intervaltab">On the Interval tab, convert distance-based "Rest" steps to "Recovery" steps</string>
+  <string name="Convert_reststep_with_type_distance_to_recoverystep_for_Advancedtab">On the Advanced tab, convert distance-based "Rest" steps to "Recovery" steps</string>
+  <string name="Export_database_to_sdcard_eg_for_upgrade">Export database to SD card (e.g . before upgrading)</string>
```
gerhardol: Remove SD card, comment about upgrade is not really needed (the database is updated, but you may want to export to be able to downgrade)

Export database to storage
Import database from storage

common/src/main/res/values/strings.xml
12people: The original made absolutely no sense to me. And, judging from the comments on Transifex, I'm not the only one.

I'm definitely open to other formulations.

common/src/main/res/values/strings.xml
12people: Okay, we can leave it out :)

common/src/main/res/values/strings.xml
gerhardol: The target pace to compare is calculated over a moving average, comparable to a window. This is not related to a frequency. The calculations have some tweaks too.
I do not have a better description right now.

PR #725, opened 2018-08-12
Design polish

12people:
I went through the app and tweaked lots little stuff -- tweaking dimensions to fit an 8dp grid, adding action bars, adding color behind the status bar, switching the two remaining Activities to AppCompatActivities (no harm done), and making the "Manage workouts" screen look more material.

Went through it on Froyo, no breaking changes, though I've run into some bugs that persist from the old version. One crashing bug--the HR Monitor screen crashes--and some cosmetic bugs: up buttons don't work as they should, button text sometimes randomly jumps into vertical orientation, the "Manage workouts" dropdown has two dropdown arrows. Just noting them down here, but, aside from the last cosmetic bug (which I don't think is necessary to deal with right now), they're not relevant to this pull request.


gerhardol:
New Lints!
https://travis-ci.org/jonasoreland/runnerup/builds/415201535#L2441

UseCompoundDrawables 
Is that fixable in a simple way?

UnusedResources
Remove ButtonTextGrey?

ContentDescription
Suppress?

12people:
Fixed!

gerhardol:
@12people did you see my PR?

gerhardol:
<string name="dialog_ellipsis" translatable="false">%s…</string>
(not yet in GitHub, only Transifex)
Will set unused resource as well on some that were commented out.
Transifex may trash these attributes though

12people:
@gerhardol The idea was to let it be translatable for e.g. RTL languages, if it is indeed different for them.

PR #737, opened 2018-10-07
History tab: Fixing layout overlay bug

12people:
The only changes here are the addition of a LinearLayout.

The other changes are simply the result of Android Studio's code reformat command (which should keep the code more readable for future uses).


gerhardol:
Please provide screenshots of changes like this, I had to look carefully until I realized that the header "Oct 2018" was hidden.

For reformat, I prefer to have them as separate commits. 
Reformat all code is an option. (I do not have any major outstanding branches.)

12people:
Okay, will keep it in mind

PR #747, opened 2019-01-10
Fix dutch menu

maxvisser:
changed two words to fix menu issue with dutch localisation (did not test)


gerhardol:
Please change this in Transifex or it will be overwritten at next update...
https://www.transifex.com/runner-up/runner-up-android/language/nl_NL/

PR #798, opened 2019-06-11
Remove Froyo product flavor from master

gerhardol:
Froyo variant crashes on SDK>=15 and fails to start on Gingerbread emulator
No compatibility is actively removed
It may be possible to get Froyo working with using older SDK, but Oreo builds must be pushed to Play before Android Q is released

Some Lint rearrangements for Latest

Froyo must be handled in a separate branch to handled separate target APIs etc


gerhardol:
Plan to merge this if no successful test reports for v2.0.0.6

PR #843, opened 2019-08-30
Update README.md

Poussinou:
Hi,

This small PR adds a link to get your app on F-Droid 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@ RunnerUp
 ========
 
 [![Join the chat at https://gitter.im/jonasoreland/runnerup](https://badges.gitter.im/jonasoreland/runnerup.svg)](https://gitter.im/jonasoreland/runnerup?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-
 [![Build Status](https://travis-ci.org/jonasoreland/runnerup.svg?branch=master)](https://travis-ci.org/jonasoreland/runnerup)
 
 Track your sport activities with RunnerUp using the GPS in your Android phone.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,16 +24,10 @@ Track your sport activities with RunnerUp using the GPS in your Android phone.
 
 * [Play Store](https://play.google.com/store/apps/details?id=org.runnerup).
 * [GitHub](https://github.com/jonasoreland/runnerup/releases). (RunnerUp 2.x versions including beta and release candidates.)
+* [F-Droid](https://f-droid.org/packages/org.runnerup.free/). This version does not contain some features due to licensing, see the F-Droid description.
 
-<!--
-* [F-Droid](https://f-droid.org/repository/browse/?fdid=org.runnerup.free). This version does not contain some features due to licensing, see the F-Droid description.
-
-[![Get it on F-Droid](get_it_on_f-droid.png?raw=true)](https://f-droid.org/repository/browse/?fdid=org.runnerup.free)
```
gerhardol: This badge should maybe be included too - it was before.

Marx2:
It's not safe to put such links in apps. Is this link visible in the app?
Google can take down app which expose such links.

sob., 31 sie 2019, 00:35 użytkownik Gerhard Olsson <notifications@github.com>
napisał:

gerhardol:
The readme is only visible on github

Poussinou:
There is a build available for download, both in the android client and on the website.

Poussinou:
If you want to keep the F-Droid button, we could also add a Google Play one like this, if you want

`[Get it on Google Play]()`

gerhardol:
Install from f-droid fails with "unknown error" for me. Cannot check log at right now. Better for anyone else? 

Maybe the list should be replaced with the buttons.

gerhardol:
And fdroid page do not have the screenshots of the app - not really this pr but just noting.

gerhardol:
I can with adb install but not from F-Droid or a local apk.
The reason is this:
Finsky  : [9587] xfg.a(19): Installation silently blocked for org.runnerup.free

Solution is to deactivate Google Play -> Play Protect -> Scanning device for security threats
Other apps are OK though

gerhardol:
> If you want to keep the F-Droid button, we could also add a Google Play one like this, if you want

Let us add two buttons instead of the list

Poussinou:
Hum I should have been more careful.

I can download the apk, but the installation was successful only at the third try. Don't know why.

Can you try multiple time, or try to install the app with another phone, to see if it's reproducible?

gerhardol:
On another phone (6.0) it was fine, with the Play Protect warning though.
On the main phone (9.0) it worked this time. Many failures previously.
I have had similar problem with private certificates. Maybe Play Protect refuses also with  "Install anyway" for the first occurrences or install once without play protect opens install?

F-Droid is quite slow right now, it can take many tries to update the applists, so maybe it is that.

For this PR, I believe install buttons should be used and list GH as an alternative.

gerhardol:
I was waiting for the Play version to be released, sorry for not merging...
Thanks

PR #850, opened 2019-10-20
speed (and)/or pace

Tiefkuehlpizze:
Hi,
I'm a long term user of this great app and I finally could make my first contribution.

I'd like to present the option to change between pace and speed (not sure, how to call it). It offers a setting to change the whole app (including graphs) to either display speed (km/h, mph) or pace (min/km, min/mi). German Translation is included.

I've never worked with Android applications before, but I'm feeling confident, that the hours full of joy inspecting the apps structure created a decent result.

### DIFF
--- a/app/res/xml/settings.xml
+++ b/app/res/xml/settings.xml
@@ -19,13 +19,26 @@
     android:layout_width="fill_parent"
     android:layout_height="fill_parent" >
 
+    <PreferenceScreen
+        android:title="@string/Unit_preference_title"
+        android:key="units_preferencescreen">
 
-    <ListPreference
-        android:key="@string/pref_unit"
-        android:persistent="true"
-        android:title="@string/Unit_preference"
-        android:entries="@array/unitEntries"
-        android:entryValues="@array/unitValues" />
+        <ListPreference
+                android:key="@string/pref_unit"
+                android:persistent="true"
+                android:title="@string/Unit_preference"
+                android:entries="@array/unitEntries"
+                android:entryValues="@array/unitValues" />
+
+        <ListPreference
+                android:key="@string/pref_speedunit"
+                android:persistent="true"
+                android:title="@string/Speed_unit_preference"
+                android:defaultValue="@string/setting_default_speedunit"
+                android:entries="@array/speedUnitEntries"
+                android:entryValues="@array/speedUnitValues" />
+
+    </PreferenceScreen>
 
     <PreferenceScreen
         android:title="@string/Sensors"
--- a/app/src/main/org/runnerup/util/Formatter.java
+++ b/app/src/main/org/runnerup/util/Formatter.java
@@ -29,7 +29,9 @@ import android.text.format.DateUtils;
 import android.util.Log;
 
 import org.runnerup.R;
+import org.runnerup.common.util.Constants;
 import org.runnerup.workout.Dimension;
+import org.runnerup.workout.SpeedUnit;
 
 import java.text.DecimalFormat;
 import java.text.SimpleDateFormat;
@@ -221,6 +223,26 @@ public class Formatter implements OnSharedPreferenceChangeListener {
                 editor.putString(key, "km");
         }
         return true;
+    }
+
+    /**
+     * Gets the user's preferred Speedunit
+     *
+     * @param context Context element to access configuration
+     * @return Configured Speed Unit (falls back to pace, if the configured value is invalid)
+     */
+    public static SpeedUnit getPreferredSpeedUnit(Context context){
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        // use either pace or speed according to the user's preference
+        String speedUnit = prefs.getString(context.getResources().getString(R.string.pref_speedunit), SpeedUnit.PACE.getValue());
+        assert speedUnit != null;   // may not happen
+        switch(speedUnit){
+            case Constants.SPEED_UNIT.SPEED:
+                return SpeedUnit.SPEED;
+            case Constants.SPEED_UNIT.PACE:
+            default:
+                return SpeedUnit.PACE;
+        }
     }
 
     public double getUnitMeters() {
@@ -461,6 +483,27 @@ public class Formatter implements OnSharedPreferenceChangeListener {
     }
 
     /**
+     * Returns either a formatted value in minutes per kilometer or kilometer per hour
+     * depending on the user's preference
+     *
+     * @param target the target format [cue or txt] and length
+     * @param meters_per_second speed in m/s
+     * @return ready to speak or display value
+     */
+    public String formatVelocityByPreferredUnit(Format target, Double meters_per_second) {
+        // use either pace or speed according to the user's preference
+        String paceTextUnit = this.sharedPreferences
+                .getString(context.getResources().getString(R.string.pref_speedunit), SpeedUnit.PACE.getValue());
+        // avoid stored null to cause jumping into the distance per time branch and default to pace
+        paceTextUnit = paceTextUnit != null ? paceTextUnit : SpeedUnit.PACE.getValue();
+        if(paceTextUnit.contentEquals(SpeedUnit.PACE.getValue())) {
+            return this.formatPaceSpeed(target, meters_per_second);
+        }else {
+            return this.formatSpeed(target, meters_per_second);
+        }
+    }
+
+    /**
      * Format pace from speed
      *
      * @param target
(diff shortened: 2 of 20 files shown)


app/src/main/org/runnerup/util/GraphWrapper.java
```diff
@@ -245,7 +249,15 @@ void emit(double tot_distance) {
             }
 
             if (avg_dist > 0) {
-                double pace = avg_time / avg_dist / 1000.0;
+                double pace = 0.0;
```
gerhardol: please rename the variable and change the check for dist to be only for pace, a similar for speed but with time

common/src/main/res/values-de/array.xml
```diff
@@ -8,6 +8,14 @@
     <item>mi</item>
     <item>km</item>
   </string-array>
+  <string-array name="speedUnitEntries">
```
gerhardol: This will be added from Transifex, so it may be overwritten (unless this is manuallt retained when merging).

common/src/main/res/values/array.xml
```diff
@@ -24,6 +24,14 @@
     <item>mi</item>
     <item>km</item>
   </string-array>
+  <string-array name="speedUnitEntries">
+    <item>Pace (Minutes per kilometer)</item>
+    <item>Speed (Kilometers per hour)</item>
+  </string-array>
+  <string-array name="speedUnitValues">
```
gerhardol: This must not be translated and be kept in the same order as SpeedUnit (dont remember exactly how this is normally done)

app/src/main/org/runnerup/util/Formatter.java
```diff
@@ -1,758 +1,804 @@
...
+     *
+     * @param target the target format [cue or txt] and length
+     * @param meters_per_second speed in m/s
+     * @return ready to speak or display value
+     */
+    public String formatSpeedByPreferredUnit(Format target, Double meters_per_second) {
+        // use either pace or speed according to the user's preference
+        String paceTextUnit = this.sharedPreferences
+                .getString(context.getResources().getString(R.string.pref_speedunit), "pace");
+        if(paceTextUnit != null && paceTextUnit.contentEquals("pace")) {
```
gerhardol: Keep pace default if paceTextUnit == null

Tiefkuehlpizze:
Thanks for the review!  Good to hear, that you like it. I don't know what to do about Transifex. Probably just wait until it's getting overwritten? :)  I'll implement the changes in the next days.

common/src/main/res/values/array.xml
Tiefkuehlpizze: I'll test it without the translation. Maybe my intention was to eager. :)

app/src/main/org/runnerup/util/Formatter.java
Tiefkuehlpizze: Good one! I haven't considered, that there could be a null value in the configuration where my default wouldn't work.

app/res/xml/settings.xml
```diff
@@ -27,6 +27,14 @@
         android:entries="@array/unitEntries"
         android:entryValues="@array/unitValues" />
 
+    <ListPreference
```
gerhardol: This should be a subsection in Unit preferences, not a new top level category

app/src/main/org/runnerup/util/Formatter.java
```diff
@@ -1,758 +1,804 @@
-/*
```
gerhardol: Please retain the attributes on the file

common/src/main/res/values/array.xml
gerhardol: They still need to be translated. The problem is that the array cannot be changed after releasing to translators

Tiefkuehlpizze: What would you expect here? I understand the potential issues between displayed text and actual value. It was done the same way with the `unitEntries` and `unitValues` arrays above my addition.

app/res/xml/settings.xml
Tiefkuehlpizze: Shifted both into a new submenu

app/src/main/org/runnerup/util/Formatter.java
Tiefkuehlpizze: Windows line endings... unfamilar

app/src/main/org/runnerup/util/GraphWrapper.java
```diff
@@ -244,22 +248,32 @@ void emit(double tot_distance) {
...
-                }
-                paceList.add(new DataPoint(tot_distance, pace));
-                hrList.add(new DataPoint(tot_distance, Math.round(avg_hr)));
-                acc_time = 0;
-
-                tot_avg_hr += avg_hr;
-                avg_pace += pace;
-                min_pace = Math.min(min_pace, pace);
-                max_pace = Math.max(max_pace, pace);
+            double speed = 0.0;
```
gerhardol: calling it speed instead of pace is not good either. speedOrPace maybe
min_pace/max_pace are incorrect derived too in the speed scenario (unsure of the effect).

Tiefkuehlpizze: Yeah there is something hidden. I've imported my read data on my device to better see the effects.
The graph tries to display the speed as time right now. I'll have another look into it.

In the meantime, I tried to find a better word to describe speed and pace together and I got the word _velocity_. Does it sound better? Should it replace everything I've labeled with _speed_?

Where is the column header you are speaking about?

gerhardol: image

velocity is OK

Tiefkuehlpizze: Got it! It's now velocity in English.
Graph is also fixed. Seems like I had a calculation issue and missed some important formatting stuff.

gerhardol: I have been to busy the last week and maybe I should not have written anything at all without being clearer...
I meant that velocity was an OK for name for the variable. I would prefer to use Pace/speed for the column header and graph. Maybe can live with that....
Some further comments too.

common/src/main/res/values/array.xml
```diff
@@ -24,6 +24,14 @@
     <item>mi</item>
     <item>km</item>
   </string-array>
+  <string-array name="speedUnitEntries">
+    <item>Pace (Minutes per kilometer)</item>
```
gerhardol: This does not work for Miles
Either update the array before displaying (I do not know how right now) so the correct unit is used or just use speed/pace

common/src/main/res/values/array.xml
gerhardol: The array cannot be changed from now on without simultaneously updating translations.

app/src/main/org/runnerup/util/GraphWrapper.java
```diff
@@ -221,57 +225,67 @@ void emit(double tot_distance) {
...
+                hrList.add(new DataPoint(0, Math.round(avg_hr)));
+                first = false;
+            }
+            velocityList.add(new DataPoint(tot_distance, velocity));
+            hrList.add(new DataPoint(tot_distance, Math.round(avg_hr)));
+            acc_time = 0;
+
+            tot_avg_hr += avg_hr;
+            avg_velocity += velocity;
+            min_velocity = Math.min(min_velocity, velocity);
```
gerhardol: This will likely not be as expected for pace, but this is not changed from before.

app/src/main/org/runnerup/util/Formatter.java
```diff
@@ -223,6 +224,26 @@ private static boolean guessDefaultUnit(Resources res, SharedPreferences prefs,
...
+    /**
+     * Gets the user's preferred Speedunit
+     *
+     * @param context Context element to access configuration
+     * @return Configured Speed Unit (falls back to pace, if the configured value is invalid)
+     */
+    public static SpeedUnit getPreferredSpeedUnit(Context context){
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        // use either pace or speed according to the user's preference
+        String speedUnit = prefs.getString(context.getResources().getString(R.string.pref_speedunit), "pace");
```
gerhardol: Use SPEED_UNIT instead of hardcoding "pace"

app/src/main/org/runnerup/util/GraphWrapper.java
Tiefkuehlpizze: I'm not even sure where these values are used or displayed.

app/src/main/org/runnerup/util/Formatter.java
```diff
@@ -233,7 +233,7 @@ private static boolean guessDefaultUnit(Resources res, SharedPreferences prefs,
     public static SpeedUnit getPreferredSpeedUnit(Context context){
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
         // use either pace or speed according to the user's preference
-        String speedUnit = prefs.getString(context.getResources().getString(R.string.pref_speedunit), "pace");
+        String speedUnit = prefs.getString(context.getResources().getString(R.string.pref_speedunit), SpeedUnit.SPEED.getValue());
```
gerhardol: "pace" is hardcoded in at least 5 other places in Formatter.java (as well as settings.xml, nothing to do there)

Tiefkuehlpizze: Done. Thanks for being patient.

gerhardol:
Merged in #855

PR #873, opened 2020-01-11
Get maximum of heart rate and use 110% for y axis of heart rate graph

vanseforge:
The maximum of the heart rate graph in the detail view was set by the graphview. I changed it to 110% of the maximum heart rate from settings. Now the graph is better scaled.

### DIFF
--- a/app/src/main/org/runnerup/util/GraphWrapper.java
+++ b/app/src/main/org/runnerup/util/GraphWrapper.java
@@ -26,6 +26,7 @@ import android.os.AsyncTask;
 import android.os.Build;
 import android.preference.PreferenceManager;
 import android.util.Log;
+import android.util.Pair;
 import android.view.View;
 import android.widget.LinearLayout;
 import android.widget.LinearLayout.LayoutParams;
@@ -103,6 +104,14 @@ public class GraphWrapper implements Constants {
         });
         graphView2.getViewport().setScalable(true);
         graphView2.getViewport().setScrollable(true);
+
+        // Get maximum of heart rate and use 110% for y axis
+        HRZones hrZones = new HRZones(context);
+        HRZoneCalculator hrZoneCalculator = new HRZoneCalculator(context);
+        int zoneCount = hrZoneCalculator.getZoneCount();
+        Pair<Integer, Integer> values = hrZones.getHRValues(zoneCount);
+        graphView2.getViewport().setMaxY(values.second*1.1);
+        graphView2.getViewport().setYAxisBoundsManual(true);
 
         hrzonesBar = new HRZonesBar(context);
     }


app/src/main/org/runnerup/util/GraphWrapper.java
```diff
@@ -104,6 +105,14 @@ public String formatLabel(double value, boolean isValueX) {
         graphView2.getViewport().setScalable(true);
         graphView2.getViewport().setScrollable(true);
 
+        // Get maximum of heart rate and use 110% for y axis
+        HRZones hrZones = new HRZones(context);
+        HRZoneCalculator hrZoneCalculator = new HRZoneCalculator(context);
+        int zoneCount = hrZoneCalculator.getZoneCount();
+        Pair<Integer, Integer> values = hrZones.getHRValues(zoneCount);
```
gerhardol: If no zone is set, this will return null, must add a check.

vanseforge:
The goal was to get a better scaled graph for the heart rate (and always the same scaling). 

I totally forgot the min. Thank you. I added also the check if a zone is defined.

ScreenHeartRatepng

app/src/main/org/runnerup/util/GraphWrapper.java
```diff
@@ -104,6 +105,19 @@ public String formatLabel(double value, boolean isValueX) {
...
         graphView2.getViewport().setScrollable(true);
 
+        // Get min-/maximum of heart rate and use 90%/110% for y axis
+        HRZones hrZones = new HRZones(context);
+        HRZoneCalculator hrZoneCalculator = new HRZoneCalculator(context);
+        int zoneCount = hrZoneCalculator.getZoneCount();
+        Pair<Integer, Integer> MaxValues = hrZones.getHRValues(zoneCount);
+        Pair<Integer, Integer> MinValues = hrZones.getHRValues(1);
+        if(MaxValues != null && MinValues != null){
+            graphView2.getViewport().setMaxY(MaxValues.second*1.1);
```
gerhardol, line 115: This should be set in complete()

gerhardol (commented):
The presentation if below the low zone just looks like a bug to users.
Screenshot_20200830-231516
This from a walking activity, not the primary use of RU, but still a use case.

If zones are not setup, the presentation is a little stretched out:
Screenshot_20200831-005953

Presenting zero values should be considered a bug. This is how nozone presentation look after changing that:
Screenshot_20200830-224746

This could be made a configurable, but RU has enough configurations already.
Therefore, I propose that the max/min are set to the zones that match them. This should give you reasonable comparable graphs but with better resolution. If lower zone is 0, use min value instead of 0, but crop max values.

PR #885, opened 2020-01-27
GPX export: Do not export  empty trkseg

### LINKED ISSUE #884: Segment simplification: GPX segments lost
Path simplification is a great new feature in v2.0.5.1. However I noticed a weird behavior when exporting to GPX. This is similar to what @dratasich reported [here](https://github.com/jonasoreland/runnerup/pull/871#discussion_r363344319) for laps. I have a recording with (too) many pauses:
Screenshot_20200127-145054_RunnerUp
I think I tapped the pause button inadvertently... When zooming in some additional pauses can be seen.

When I export to GPX without simplification, RunnerUp creates multiple segments, which is ok. Here is how it looks like in a GPX viewer:
before

When I activate simplification on GPX export, runner up still creates segments (unlike what I thought was implemented [here](https://github.com/jonasoreland/runnerup/pull/871/commits/2c4daeede32f2deab157faead059b32bf563df61)), and some of them are either empty (\<trkseg>\</trkseg>) or have a single \<trkpt>, so not visible in GPX viewers:
after

I don't think this is correct. It is good to use GPX segments to split track when there are pauses. But simplification should then only happen *per segment*, and each segment should at least keep its first and last points, to properly keep the beginning and end of the segment.

gerhardol:
Fixes #884 

WIP

### DIFF
--- a/app/src/main/org/runnerup/export/format/GPX.java
+++ b/app/src/main/org/runnerup/export/format/GPX.java
@@ -176,20 +176,15 @@ public class GPX {
                 simplifier.getNoisyLocationIDs(mDB, activityId) :
                 new ArrayList<>();
         // Tracks with time gaps may show as unconnected, mostly the same as this setting
-        final boolean useTrkSeg = (simplifier == null);
+        final boolean useLapTrkSeg = (simplifier == null);
 
-        if (!useTrkSeg) {
-            mXML.startTag("", "trkseg");
-        }
+        // number of GPS points in current track segment
+        int segmentPoints = 0;
         while (lok) {
             if (cLap.getFloat(1) != 0 && cLap.getLong(2) != 0) {
                 long lap = cLap.getLong(0);
                 while (pok && cLocation.getLong(0) != lap) {
                     pok = cLocation.moveToNext();
-                }
-                boolean hasPoints = false;
-                if (useTrkSeg) {
-                    mXML.startTag("", "trkseg");
                 }
                 if (pok && cLocation.getLong(0) == lap) {
                     long last_time = 0;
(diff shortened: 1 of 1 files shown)


opoto:
I'm not sure it answers all of my comments in #884, but at least it should avoid creating empty segments, which is good!
Thanks

gerhardol:
This will not handle removing points around trkseg start/stop for sure.
If the simplification was very aggressive, the other problems you saw will be handled.
To test this will probably take days for me.

I appreciate formal reviews...

opoto:
I will review and test on my side. But it will take some days too.
Thanks

opoto:
I confirm that with this update, using the same recorded data and same GPX simplification parameter as #884, all generated \<trkseg> have more than 1 \<trkpt>. No more empty or single point segment. Result is closer to original path, which is good:
Capture d’écran 2020-01-28 à 23 58 31
I still think it is weird that the resulting path covers some area where the original record was in pause, but this is probably a wider discussion beyond the scope of this PR.

gerhardol:
> I still think it is weird that the resulting path covers some area where the original record was in pause, but this is probably a wider discussion beyond the scope of this PR.

How much "extra"?
The last known location is logged when pausing, but that should max add the GPS sampling time (default a second). Otherwise too little could be covered.

However, I believe that what you see here is that since pauses are not preserved, the 'single' points stretches from one segment to another and thereby covers the pause. There are simply not enough points to describe the segments.

The simplification must be done by segment as you suggested for this to be completely correct.
(A hack would be to reinsert the points around the pauses...)

gerhardol:
Updated slightly, to avoid (possible) empty segment in the end of a track

PR #895, opened 2020-02-23
Fix path simplification:

opoto:
The "optimization" I introduced to skip simplification when segment only contained 2 locations was discarding those segments... I removed this optimization, which was not significant anyway.


gerhardol:
seem OK

gerhardol:
So this is no longer needed?

opoto:
Correct, this PR can be discarded as the fix was also included in #894

PR #896, opened 2020-02-28
Android Studio 3.6.2

gerhardol:


gerhardol:
AS updated since this was submitted (I hold off merging to allow #894 to be ready first).

Issue #897, opened 2020-04-01
Discontinued upload servers


opoto:
Runner Up still offers to upload recorded tracks to servers/accounts which are not working anymore:

- Digifit: https://www.digifit.com
- FunBeat: https://www.funbeat.se
- Google Plus: https://plus.google.com
- Nike plus: https://nikeplus.nike.com
- RunnerUp LIVE: https://weide.devsparkles.se/Demo/Map
- RunningFreeOnline: https://www.runningfreeonline.com

They should be removed from Runner Up in order to simplify account preference UI, avoid user clicking on it and being stalled or get an error, reduce app size, and reduce source code to maintain.

gerhardol:
I have not removed services that are not working as someone may use the template for a modified site and fix it. RU is not owned by me either.
So I have disabled but not removed services that are not working.
For these, LIVE may be revived if you setup that webserver (I have not tried it), the others are safe to remove (there are more candidates).

So please go ahead (maybe leave LIVE).
Garmin, Facebook will very likely never work (with the current APIs).
Runtastic and MapMyRun are unknown
GoogleFit, jogg.se should be fixable what I know

RunningAhead need to be investigated too (site still exists, so just disabling for now if not working).
RunKeeper API is deprecated and a PR to disable already exists.

Too bad that Tapiiriik.com is not fully working, that is what I added Dropbox for...

opoto:
Fixed by #898

PR #898, opened 2020-04-03
Remove deprecated accounts

opoto:
As discussed in #897, removed some accounts:
- Digifit
- FunBeat
- Google Plus (moved OAUTH2 code to GoogleFit)
- Nike plus
- RunningFreeOnline

As suggested, kept RunnerUp LIVE in case someone reinstalls the server.
Did not dare remove Garmin, Facebook, Runtastic, MapMyRun, GoogleFit, and jogg.se, because these services still exist and someone may fix the synchronizers to bring them back to life.


opoto:
You're right, I missed that.
What about deleting DB entries on the fly during account listing, when seeing that the synchronizer does not exist anymore?

gerhardol:
> What about deleting DB entries on the fly during account listing, when seeing that the synchronizer does not exist anymore?

That is a variant of the current "always run" that disables accounts.
This will clear the status about uploaded activities.
Can you test how this behaves?

A variant is to set the account status to the new status "removed" at strtup and avoid displaying them in the account list.

It is hard to make such changes on your own, I have struggled for several years now without a discussion partner...

gerhardol:
Follow up on RunningAHEAD: the service is flaky, but not dead. Let us leave it as is

opoto:
I fully understand the difficulty to maintain this project alone, and I congratulate you for the work you do!
The on-the-fly approach I imagined seems to be too complex: there are many (complex) execution paths to read the accounts, and I'm sure it would miss a scenario. So DB upgrade seems to be the more reasonable way to remove these accounts.

gerhardol:
A new db version adds a complexity when developing, I prefer to not do that.
Also, dbversion is not strictly required here as the format is not changed.

I was going to suggest to set ENABLED=2 in dbhelper, and add ENABLED!=2 to AccountListActivity.java, ManageWorkoutsActivity.java,
but deleting in the same way could be enough.

opoto:
Deletion on the fly seems to work, I tested though 3 flows:
- open activity in history (although my activities did not refer to deleted accounts)
- feed
- accounts in preferences

gerhardol:
Seem OK, will try tomorrow

app/src/main/play/listings/en-US/full-description.txt
```diff
@@ -7,20 +7,19 @@ Track your sport activities with RunnerUp using the GPS in your Android phone:
...
 * Pebble support
 
-Upload your activities to a variety of running sites with a single click:
+Upload your activities to a variety of external applications with a single click:
 
 * Strava
 * Runalyze
 * Endomondo
-* RunKeeper
-* RunningAHEAD
```
gerhardol, line 23: Not dead yet, so keep, just less promoted

opoto: done

gerhardol:
The TripleX doc (basis for F_Droid and Play) should be synced. Planned to do this myself

opoto:
I don't know what is the TripleX doc, and did not find anything with that name in the repo...

gerhardol:
app/src/main/play

Triple-T...
https://github.com/Triple-T/gradle-play-publisher

opoto:
Updated and squashed

gerhardol:
Thanks!

PR #654, opened 2018-04-23
Account list redesign

12people:
Pushing a the redesign of the account list.

In the process of doing the redesign, I discovered that a number of the logos blatantly went against companies' guidelines on when to use their logos, which makes me think they were used without permission. In order to stay safer legally, I replaced the offending logos with simple avatars—their first letters inside circles.

The "Show disabled accounts" button has been moved to the bottom for better discoverability. The grid and table views were merged into one view featuring the name of the service as well as switches for turning upload or social on or off.

screenshot_20180424-010217

PR #900, opened 2020-04-09
Update and add missing icons for enabled accounts

opoto:


gerhardol:
For use of icons, @12people removed previous versions of them in #654 as the use was not free.
Have you got other information?

opoto:
I'm not legal expert, and I don't have detailed info.
What I used is:
- Strava: 
I used https://www.freepng.fr/png-y8z4t8/, which is free for no commercial usage.
Official Strava logos are here: https://blog.strava.com/fr/press/assets/
API documentation looks very restrictive with the icon to add, it would not fit to RU account UI: https://developers.strava.com/guidelines/
- Dropbox:
I used the official logo from https://www.dropbox.com/sh/cclw5exikk03wmv/AAB8jpudVftt1e2m-D_CS9QMa/Logos?dl=0&subfolder_nav_tracking=1
I think its usage in RU is compliant with https://www.dropbox.com/branding
- Runalyze:
I used https://runalyze.com/favicon.ico, for which I see no restrictions
- Endomondo:
I used the Wikipedia icon: https://en.wikipedia.org/wiki/File:Endomondo_Logo.png
Actually it may not be up to date, and license is not so clear. So may be this is not ok to use this one. But I could not find any official branding/logo usage recommendations on endomondo web site.
- RunKeeper: I took the app icon from Google Play: https://play.google.com/store/apps/details?id=com.fitnesskeeper.runkeeper.pro
I added transparency and inverted the colors. I did not find any official branding/logo usage recommendations on runkeeper web site.

So all in all:
- I don't see any problem with Dropbox and Runalyze
- Strava and RunKeeper is not official logos, but not sure this is an issue for RU
- Endomondo might need to be updated

What do you think?

gerhardol:
From the app perspective, an unofficial icon is OK.
The way the accounts are presented, I find it obvious that RU connect to these 3rd part services, not pretending to be part of them or even be endorsed.

The problem is if the owners request that the logos and usage should be removed, just to retain their ownership.
Google can be fickly to reactivate apps if they get such a request. (2.0 should be added to Play, even if it may have to be added to a new account.)

> Strava

Strava guidelines mandates a "Connect with Strava" icon, so we have to use that and should change (to something slightly different).
https://developers.strava.com/guidelines/

> Dropbox

I have tried to confirm to that page, good that you agree
(not changed here)

> RunAlyze

For RunAlyze I got an OK to use the current icon. That icon is not usable as a favico, but I find it OK in the app, easier to relate to the logo.

> Endomondo

The API is unofficial, but has been stable for years so it is at least tolerated.

> RunKeeper

RunKeeper API is retired, to be deactivated and should not be updated.

gerhardol:
> Strava guidelines mandates a "Connect with Strava" icon, so we have to use that and should change (to something slightly different).
https://developers.strava.com/guidelines/

RU start page and TripleX description (used for F-Droid, should be used in Play too) need to be updated too (probably separate PR)

opoto:
Ok, reverted changes on RunKeeper. Should I "disable" it in this PR, or do you want to use #701?
Maybe I can update README and full_description in #898?
For the "Connect to Strava" official icon, it cannot fit in the account list screen. So I just replaced with official Strava icon from https://blog.strava.com/fr/press/assets/. If you think RU has to include the dev icon, then what could be done is introduce a "large icon" for each synchronizer (defaulting to the existing one) to be used in the account edition screen.

opoto:
Wiki should also be update
https://github.com/jonasoreland/runnerup/wiki/Synchronization-with-external-providers

gerhardol:
I am not sure any of this can be included, trademark violations must be avoided. Sorry...

For Strava, the icon should be updated according to the Strava requirements, the press material cannot be used in the app.
Need to see what other apps are using if the update is tolerated or not
This is the service I care mostly about

For Dropbox, if this is the 'logo' from the API page I believe this is OK to change (I used the app icon). But it must not be changed from the original.

RunAlyze should have the existing icon in my view

RunKeeper should just be kept as is, removed when no longer working.

Endomondo is unofficial, keep as is

Update the readme here or in #898 if you prefer

opoto:
Ok, so maybe we should just discard this PR then

gerhardol:
> Ok, so maybe we should just discard this PR then

Maybe Strava, Dropbox

opoto:
OK, just kept Strava and Dropbox

gerhardol:
Endomondo is added and Runalyze updated still.
I have not found so many other apps with these icons.
Runalyze: IpBike and LocusMap has the favico icon too (MyRunningApp has none).
Stgrava: IpBike has no primary icon, but the second page has the required "Connect with Strava". Other apps has a variant of this app.
Endomondo: I have not found other apps with upload or icons, the interface is unofficial, so let us keep it - I do not see that this is negative for Endomondo
Dropbox: Some variants are used in apps, the update should be allowed too.

So let us keep as is
I will squash merge unless you squash

opoto:
Actually it is already squashed

gerhardol:
master was merged to the feature branch, that should never bw done...
Squashed was OK
Thanks

PR #901, opened 2020-04-13
Fix RunnerUpLive on Android Oreo

matgoebl:
RunnerUpLive did not work on my Android Oreo (8.1, API level 27):

-  it was not possible to activate live logging, because the config item pref_runneruplive_active was commented out. I re-activated it and moved it to a suitable submenu.
- `startForegroundService()` in LiveService crashed. Up to API level 25 `startService()` is used. Using `startService()` up to API level 27 works.

Crash log:
`ActivityManager: Bringing down service while still waiting for start foreground: ServiceRecord{bea663d u0 org.runnerup/.export.RunnerUpLiveSynchronizer$LiveService}`
`AndroidRuntime: Shutting down VM`
`AndroidRuntime: FATAL EXCEPTION: main`
`AndroidRuntime: Process: org.runnerup, PID: 8618`
`AndroidRuntime: android.app.RemoteServiceException: Context.startForegroundService() did not then call Service.startForeground()`

### DIFF
--- a/app/res/xml/settings.xml
+++ b/app/res/xml/settings.xml
@@ -161,6 +161,13 @@
             android:title="@string/log_extended_gps_title"
             android:summary="@string/log_extended_gps_summary" />
 
+        <CheckBoxPreference
+            android:defaultValue="false"
+            android:key="@string/pref_runneruplive_active"
+            android:persistent="true"
+            android:title="@string/Enable_RunnerUpLive"
+            android:summary="@string/Note_you_need_to_connect_to_the_account_too"/>
+
         <PreferenceCategory
             android:key="@string/path_simplification"
             android:title="@string/path_simplification">
@@ -376,13 +383,6 @@
                 android:title="@string/Smooth_pace_filters" />
         </PreferenceCategory>
     </PreferenceScreen>
-
-    <!--CheckBoxPreference
-        android:defaultValue="false"
-        android:key="@string/pref_runneruplive_active"
-        android:persistent="true"
-        android:title="@string/Enable_RunnerUpLive"
-        android:summary="@string/Note_you_need_to_connect_to_the_account_too" /-->
 
     <PreferenceScreen
         android:key="advanced_preferencescreen"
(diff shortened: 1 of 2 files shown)


gerhardol:
First: What RunnerUp Live service do you use?

For disabling, this was done as there was no known service and some users were confused. Can this be described somehow?

For the startService() workaround: Do you have any references?
What I find is workarounds with 5s activation, like https://stackoverflow.com/questions/55894636/android-9-pie-context-startforegroundservice-did-not-then-call-service-star/56338570#56338570

matgoebl:
I use the RunnerUp Live connection to send the data to my private server, it has just a HTTPS endpoint that takes lat/lon from the POSTed JSON and records my position. There have been discussions mentioning alternative live services - I did not try them yet.
This Live option is useful to connect it to any backend service (that is able to pick the data from the posted json structure).

If the goal was to deactivate it, why was only the checkbox to activate the service removed? Still being able to configure an account is misleading to me.

For the `startService()` I read at https://developer.android.com/guide/components/services
"If your app targets API level 26 or higher, the system imposes restrictions..". The app specifies `targetSdkVersion` = 28, but my device is API 27 - I suspect, it has not yet the named restrictions.

As stated in https://developer.android.com/reference/android/content/Context#startForegroundService(android.content.Intent) the correct way is to call `startForeground(int, android.app.Notification)`. But do we really want another notification for every http request?

The page you mentioned discusses android 9, my proposal is only a fix for android 8.1.

gerhardol:
> If the goal was to deactivate it, why was only the checkbox to activate the service removed? Still being able to configure an account is misleading to me.

There was no known implementation, the option was considered confusing, this was just removed from the GUI, it should have been possible to edit the settings backend if this was incorrect.

> For the `startService()` I read at https://developer.android.com/guide/components/services
> "If your app targets API level 26 or higher, the system imposes restrictions..". The app specifies `targetSdkVersion` = 28, but my device is API 27 - I suspect, it has not yet the named restrictions.

I do not have anything to test against, can you check with more devices?
 
> As stated in https://developer.android.com/reference/android/content/Context#startForegroundService(android.content.Intent) the correct way is to call `startForeground(int, android.app.Notification)`. But do we really want another notification for every http request?

I expected that the general notification is successful
 
> The page you mentioned discusses android 9, my proposal is only a fix for android 8.1.

As both are >= API 26, I expected them to work the same

As LIVE is not likely used much, I expect that this is added

Uatschitchun:
With Locus Map I use https://github.com/Uatschitchun/Latrine-live to expose location to a server and track the path. Maybe this would allow for the "Live" function to live on?

gerhardol:
There was a change to foreground services in Android 26. I tried to conform to that but had nothing to test with.
https://developer.android.com/about/versions/oreo/background

PR #902, opened 2020-04-13
Remove useless/duplicate simplification in GPX export

opoto:
(my fault, should have been removed when introducing PathCursor)

### DIFF
--- a/app/src/main/org/runnerup/export/format/GPX.java
+++ b/app/src/main/org/runnerup/export/format/GPX.java
@@ -29,7 +29,6 @@ import org.runnerup.workout.Sport;
 import java.io.IOException;
 import java.io.Writer;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


opoto:
I also wonder what is `useLapTrkSeg` for... Using GPX track segments for laps is ok, and I don't see why this should not happen when simplifying (but still happen when not simplifying!). Did I miss something?

gerhardol:
> I also wonder what is useLapTrkSeg for... Using GPX track segments for laps is ok, and I don't see why this should not happen when simplifying (but still happen when not simplifying!). Did I miss something?

GPX segments are handled similar to pauses in some GPX viewers, as laps in others. trkseg is therefore only added with the accuracy extension, that logs and export extra data (sats, bearing etc - could add signalstrength andsimilar too). The extensions could be used by a filter to do something similar to the simplification, only better as more data is available. The GPX export is for external programs. trkseg could be separated, butRU has already too much settings.

PR #906, opened 2020-04-19
antplus p.b6

gerhardol:
R8 ProGuard

This reportedly solves problem with ANT+ and Android 10
No possibility to test with Android 10 though, only a Android 7 phone


gerhardol:
The only phone I have tested with has Android 7, I have no access to Android 10 with ANT+. Are there any ANT+ users?

Issue #327, opened 2015-09-20
Owncloud upload


acht:
Hallo,

an automated upload to owncloud (to a preselected folder as gpx with name based on time) would be a nice feature.

Thanks

dradux:
@acht I agree, this would be a great new feature.

tbdal:
@acht i totally agree, too. maybe there is even a way to sync it with the calendar?

acht:
More general would be a integration with the standard android calender (which can be synced by the owncloud app) or a connection to a standard webdav/caldav calender (e.g. a owncloud server). But this would be an other feater then the file upload.

jonasoreland:
why could you sync with a calender ? what does that mean ?

/Jonas

tbdal:
before using runnerup i was used to create a manual entry in my android calender looking like "running, x km, hf yyy" place "zzz", which is synced via caldav to owncloud.
this was helpful to get an overview of the training, without having the need to sync with an external service but having all data on my own systems.
my vision would be to have a combined sync with one click: a) upload gpx to owncloud & b) entry in the calendar or even having an owncloud app which allows you to use the standard calendar&file system to display training over time

Vebryn:
:+1:

speendo:
+1 ;-)

31416r:
+1 :-)

PR #909, opened 2020-04-20
adds uploads to WebDav

ticevi:
First try in adding WebDav Support,
Upload is working with nextcloud but needs more testing.

fixes https://github.com/jonasoreland/runnerup/issues/327

### DIFF
--- a/app/src/main/org/runnerup/db/DBHelper.java
+++ b/app/src/main/org/runnerup/db/DBHelper.java
@@ -47,6 +47,7 @@ import org.runnerup.export.RunnerUpLiveSynchronizer;
 import org.runnerup.export.RunningAHEADSynchronizer;
 import org.runnerup.export.RuntasticSynchronizer;
 import org.runnerup.export.StravaSynchronizer;
+import org.runnerup.export.WebDavSynchronizer;
 import org.runnerup.util.FileUtil;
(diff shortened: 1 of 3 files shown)


gerhardol:
Any NextCloud service that works?
I cannot get the one I signed up to working, any hints on the URL?

If this is experimental, it could be disabled by default.
As this is based on existing uploaders, I see no problem.

ticevi:
The Url for WebDav with nextcloud should look similar to this: https://%domain%/remote.php/dav/files/%username%/%folder%
WebDav does not support to create more than one folder at once. -> the base folder must exist

gerhardol:
When I tried to upload files yesterday, the uploader did not report any errors, but the file did not appear in the NextCloud implementation. I did not debug it then, but today it worked.
Not much error handling - similar to the existing though, maybe a fluke.
It would be nice to add the link to the file as an external ID (similar should be added for DropBox). With both both tcx/gpx, one of them should be selected.

gerhardol:
Thanks
Could not recreate the upload problem
I considered adding references to the uploaded file, but that is a later enhancement

ticevi:
Thanks for merging 👍
I am currently busy with work and uni. When I have some spare time I will look into adding the references und doing some code cleanup for a more robust upload

PR #932, opened 2020-07-01
Include format when creating a syncronizer

### LINKED ISSUE #931: FIXED: WebDAV not synchronizing when started manually
**This issue is for documentation to prevent multiple reports. I already fixed this issue.**

**Type:** BUG

**Problem:**
When you do not have an internet connection the activity will not get synchronized. There was a problem, that the format (tcx / gpx) was not correctly passed to the WebDavSyncronizer, leading to a success without uploading the file. This issue occured on manual sync only.

**How to reproduce:**
Start an activity and save it with no internet connection. Connect your internet and go to accounts. Select WebDav, choose upload and manually select the previous activity. The Upload will look like it's succeeded and disappears from the unsynced list but there will be no file on your WebDav.

**Related Content:**
adds uploads to WebDav #909
Include format when creating a syncronizer #932 (FIX)

DeveloperOl:
Resolves #931

### DIFF
--- a/app/src/main/org/runnerup/export/SyncManager.java
+++ b/app/src/main/org/runnerup/export/SyncManager.java
@@ -147,7 +147,7 @@ public class SyncManager {
     @SuppressWarnings("UnusedReturnValue")
     public long load(String synchronizerName) {
         String from[] = new String[] {
-                "_id", DB.ACCOUNT.NAME, DB.ACCOUNT.AUTH_CONFIG, DB.ACCOUNT.FLAGS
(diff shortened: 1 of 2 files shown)


DeveloperOl:
I tested the WebDav account only. Please test other accounts, too before merging.

gerhardol:
The correction is ensuring that DB.ACCOUNT.FORMAT is included in all queries so it is available for synchronizers.
I would prefer to use a common list instead, to ensure that this does not happen again.

FORMAT is used in few synchronizers, only File and Dropbox, the handling is similar so both the problem and the correction should be the same.

gerhardol:
WebDAV/File/Dropbox should behave the same her.
There are some issues with refreshing, I do not see the problem always in older releases, but the change should be OK.

gerhardol:
Thanks!

PR #937, opened 2020-07-06
Add Chinese(zh-CN) support

chenzaichun:
Add simplified chinese translation.


app/res/values/array.xml
```diff
@@ -57,6 +57,7 @@
     <item>Suomi</item>
     <item>Svenska</item>
     <item>Türkçe</item>
+    <item>简体中文</item>
```
gerhardol, line 60: This list is intended to be in "alphabetic" sort order with local names. Is "Chinese" supposed to be sorted last like this?

app/res/values/cues.xml
```diff
@@ -16,6 +16,7 @@
   ~  along with this program.  If not, see <http://www.gnu.org/licenses/>.-->
   <string name="cue_activity">activity</string>
   <string name="cue_interval">interval</string>
+  <string name="Cue_intervall_m">interval</string>
```
gerhardol, line 19: why this string?
(may be an old string)

gerhardol:
The Chinese translation in Transifex is empty, so the translations can be pushed (I have not used that function though).
But zh and zh-CN is identical, so only one of them is really needed (I suggest zh).
In addition a squash and rebase is needed. 
I can do that, but unless you have allowed maintainer changes, it must be done in a PR I create.

app/res/values/array.xml
chenzaichun: It's fine.

app/res/values/cues.xml
chenzaichun: May be an old string, I translated it last year. I use RunnerUp everyday, so I thinks maybe there's someone would like it.

There're some other zh format, like zh-TW, zh-HK, which use traditional Chinese. Maybe someone could translate it.

gerhardol:
What to keep of zh or zh-CN?
There will be a difference for users of Android before 7 from what I can see in
https://developer.android.com/guide/topics/resources/multilingual-support

using zh, those that has traditional get simplified instead of English which is probably a good thing
But as mixing traditional and simplified will likely look weird it may be better to keep zh-CN in Transifex

This can be changed (as is done for sv-SE in Transifex and sv in RU).

So unless you have other suggestions @chenzaichun I will do the following:
 * Squash and rebase
 * remove 'zh'
 * Push zh-CN to Transifex

chenzaichun:
@gerhardol sure, I think keep zh-CN is acceptable right now. I'll add translation for traditional chinese later. Thanls.

gerhardol:
Translations pushed: https://www.transifex.com/runner-up/runner-up-android/language/zh_CN/

There are a few new strings to translate

I pushed a cleaned up commit to https://github.com/gerhardol/runnerup/tree/chenzaichun/chinese-support
Either reset your change to the latest commit, or I will open a separate PR (I cannot push to this PR)

chenzaichun:
@gerhardol I had added missing translation in transifex, but I don't know what can I do next. Could you please help? Should I add the missing translation to the xml and push to the PR again?

gerhardol:
Register in transifex and up date there

chenzaichun:
already updated in transifex. :)

gerhardol:
Can you make sure I can push to the branch (allow maintainers), then I will pull from Transifex to my squashed branch, push to this branch and merge

chenzaichun:
@gerhardol I reset the branch to the commit you squashed, and you could merge and push to this branch. Thanks.

gerhardol:
Thanks

I did a fixup to Transifex zh for two arrays, pulled the updates in another PR (also merged)

chenzaichun:
Thanks. :)

PR #966, opened 2020-09-03
Don't crash on no TTS available

### LINKED ISSUE #958: App crashes when no TTS engine present
Definitely an edge case but, the app crashes when android does not have a TTS engine enabled (e.g. the default LineageOS with F-Droid setup).

TheLastProject:
Fixes #958.

I would love your feedback on this.

### DIFF
--- a/app/src/main/org/runnerup/view/StartActivity.java
+++ b/app/src/main/org/runnerup/view/StartActivity.java
@@ -22,6 +22,7 @@ import android.app.NotificationManager;
 import android.content.BroadcastReceiver;
 import android.content.ComponentName;
 import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.ServiceConnection;
@@ -76,6 +77,7 @@ import org.runnerup.tracker.GpsInformation;
 import org.runnerup.tracker.Tracker;
 import org.runnerup.tracker.component.TrackerCadence;
 import org.runnerup.tracker.component.TrackerHRM;
+import org.runnerup.tracker.component.TrackerTTS;
 import org.runnerup.tracker.component.TrackerWear;
 import org.runnerup.util.Formatter;
 import org.runnerup.util.SafeParse;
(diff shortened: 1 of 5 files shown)


app/src/main/org/runnerup/workout/feedback/RUTextToSpeech.java
```diff
@@ -90,8 +90,14 @@ private String getId(String text) {
...
     }
 
+    public Boolean isAvailable() {
+        return textToSpeech != null;
+    }
+
     @SuppressWarnings("UnusedReturnValue")
     int speak(String text, int queueMode, HashMap<String, String> params) {
+        if (!isAvailable())
+            return 0;
```
gerhardol, line 100: curly brackets

app/src/main/org/runnerup/workout/feedback/RUTextToSpeech.java
```diff
@@ -135,6 +141,9 @@ int speak(String text, int queueMode, HashMap<String, String> params) {
      */
     private int speakWithMute(String text, int queueMode,
             HashMap<String, String> params) {
+        if (!isAvailable())
+            return TextToSpeech.ERROR;
```
gerhardol, line 145: curly brackets

app/src/main/org/runnerup/view/StartActivity.java
```diff
@@ -602,23 +604,45 @@ private Workout prepareWorkout() {
...
-                mGpsStatus.stop(StartActivity.this);
-
-                // unregister receivers
-                unregisterStartEventListener();
-
-                // This will start the advancedWorkoutSpinner!
-                mTracker.setWorkout(prepareWorkout());
-                mTracker.start();
+                if (!mTracker.isComponentConnected(new TrackerTTS().getName())) {
+                    createNewNoTtsAvailableDialog();
```
gerhardol, line 641: Why not start directly and put the popup in front without waiting?
As this is handled, the user can make decision to correct this next time instead.

gerhardol (commented):
Looks good!
Have not run it, but I would be surprised if it creates problems if a tts is available
Please squash too (I can do that when merging, but better to adjust the messages manually)

app/src/main/org/runnerup/view/StartActivity.java
gerhardol: ok

TheLastProject:
Squashed it.

>Have not run it, but I would be surprised if it creates problems if a tts is available

I have tested it both without and after manually installing a TTS and the messages only trigger when there is no TTS available.

gerhardol (approved):

gerhardol:
Thanks!

TheLastProject:
Thank you for explaining, I'll try to watch out for that if I contribute more in the future :)

DIWesser:
Thanks for sorting this out, y'all.

PR #970, opened 2020-09-10
Disable TTS dialog on run start

TheLastProject:
I've noticed this often triggers the first workout after app launch. Missed this during testing it seems. I recommend disabling this check for now as I will try to figure out a better way to fix this.

### DIFF
--- a/app/src/main/org/runnerup/view/StartActivity.java
+++ b/app/src/main/org/runnerup/view/StartActivity.java
@@ -637,11 +637,15 @@ public class StartActivity extends AppCompatActivity
     private final OnClickListener startButtonClick = new OnClickListener() {
         public void onClick(View v) {
             if (mTracker.getState() == TrackerState.CONNECTED) {
+                /* FIXME: Triggers too often on the first run since app start even if TTS exists
                 if (!mTracker.isComponentConnected(new TrackerTTS().getName())) {
                     createNewNoTtsAvailableDialog();
                 } else {
                     startWorkout();
                 }
+                */
+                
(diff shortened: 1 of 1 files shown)


app/src/main/org/runnerup/view/StartActivity.java
```diff
@@ -637,11 +637,15 @@ private void startWorkout() {
     private final OnClickListener startButtonClick = new OnClickListener() {
         public void onClick(View v) {
             if (mTracker.getState() == TrackerState.CONNECTED) {
+                /* FIXME: Triggers too often on the first run since app start even if TTS exists
```
gerhardol, line 640: Any possibility for a more permanent fix, or should the snippet be removed and comment changed?

TheLastProject: I'm definitely planning to try to fix this more properly without false positives but I don't think I'll be able to in the coming few days. I've been quite busy and testing the app has been a pain because I can't get GPS indoors and haven't figured out yet how to bypass the GPS searching so I can test more easily. This patch is mostly because there already released a version with this regression and I considered it rude to not try to at least undo the regression as soon as possible while I look for a more permanent better fix

gerhardol: OK, then I will include this in next release, probably this weekend.

Have you tried fake GPS apps like https://play.google.com/store/apps/details?id=fr.dvilleneuve.lockito

gerhardol (approved):

PR #972, opened 2020-09-12
transifex ru

gerhardol:


gerhardol:
ef7c45aa525ab116cc5968126e729efe0e06e567 in 2.2.5
Something in the db handling is incorrect, these changes are just hiding the exceptions but may give other errors. Hopefully improving the user experience still.

PR #974, opened 2020-09-22
Add workflow for Update Gradle Wrapper Action.

cristiangreco:
Hey there 👋, first of all thanks for your work on RunnerUp!

I've got a suggested change: would you be willing to use [this GitHub Action](https://github.com/gradle-update/update-gradle-wrapper-action) to automatically keep Gradle Wrapper updated to latest release?

**What does "Update Gradle Wrapper Action" do?** It can be configured to run at scheduled intervals (e.g. daily or weekly) and will check whether the Wrapper script in the repo is up-to-date to the latest Gradle release: in case a new Gradle version is available, it will create a PR to update the Wrapper. And that's it!

**Why is that a good thing?** Well, first of all it alleviates the chore of manually updating the Wrapper, as you got a task that keeps track of new Gradle releases for you! More importantly, it boosts security around the Wrapper update and usage processes: this actions verifies that the `gradle-wrapper.jar` file has not been tampered with (uses checksum comparison), and it sets the `distributionSha256Sum` property so that the new Gradle binary itself will be verified locally upon download.

**Where can I find more about?** The [README](https://github.com/gradle-update/update-gradle-wrapper-action/blob/master/README.md) contains quite detailed information!

In this PR I propose adding a new workflow which runs the action every day at midnight (but feel free to adjust the frequency as you prefer). I've verified it works correctly in my fork of the repo, and you can [see here](https://github.com/gradle-update/runnerup/pull/1) how the generated PR looks like.

The action is under active development, you can have a look at the [list of inputs](https://github.com/gradle-update/update-gradle-wrapper-action#action-inputs) currently supported. There's new features coming up soon and if you'd like to request any particular change just let me know!

I'd love to see the action used by RunnerUp as such a popular project would definitely help me spread the word, but I also genuinely hope you can find this useful. Would love your feedback! ❤️

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -1,0 +1,17 @@
+name: Update Gradle Wrapper
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  update-gradle-wrapper:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Update Gradle Wrapper
+        uses: gradle-update/update-gradle-wrapper-action@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}


gerhardol:
Thanks for your effort and the PR, but with an Android Studio project, the gradle version depends on the Android gradle plugin that depends on the Ansdroid Studio version, so gradle cannot be updated just like that.

cristiangreco:
@gerhardol thanks for your feedback!

So, given runnerup is on AGP 4.0.1 I think the update to the latest Gradle release (6.6.1) should technically be ok, correct? However I see your point and am going to close this PR as the action is not going to be useful for you as-is.

I'm collecting feature requests and would love to hear your opinion on this: do you think it'd make sense for the action to auto-detect AGP on android repos and propose Gradle updates based on this [compatibility table](https://developer.android.com/studio/releases/gradle-plugin#updating-gradle)?

gerhardol:
> So, given runnerup is on AGP 4.0.1 I think the update to the latest Gradle release (6.6.1) should technically be ok, correct? 

I was not aware of the + compatibility, so it should be correct
But as long as you follow AS updates, you are on a fairly new version

> However I see your point and am going to close this PR as the action is not going to be useful for you as-is.

Yes, likely what should be done

> I'm collecting feature requests and would love to hear your opinion on this: do you think it'd make sense for the action to auto-detect AGP on android repos and propose Gradle updates based on this compatibility table?

The most useful action would probably be to update AS/AGP and let that update the gradle wrapper...

At work (where GH is not used though), this action would be useful for some projects (others are locked down by the vendors). But there is still some compatibility issues between gradle versions (deprecated functionality) and most repo uses a gradle plugin so the benefit is not that high.

PR #976, opened 2020-10-01
Remove commented out code

TheLastProject:
I know I said in #970 that I was planning to create a more permanent fix, and I was planning on looking into this this weekend. Life has been quite busy in the mean time.

Sadly, the /e/OS update I got for my phone just today adds a TTS engine by default. As a system app and without root access, I can no longer remove the TTS and thus sadly can no longer test the no-TTS situation.

I apologize for making a promise I ended up not being able to keep, but I really hadn't expected this to happen. So I'm at least making this small PR to remove the obsolete code.

Again, my apologies.

### DIFF
--- a/app/src/main/org/runnerup/view/StartActivity.java
+++ b/app/src/main/org/runnerup/view/StartActivity.java
@@ -637,14 +637,6 @@ public class StartActivity extends AppCompatActivity
     private final OnClickListener startButtonClick = new OnClickListener() {
         public void onClick(View v) {
             if (mTracker.getState() == TrackerState.CONNECTED) {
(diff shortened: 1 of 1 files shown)


gerhardol:
No need to apologize, thanks for the efforts.
We all have limited time.
If you believe that you will look at this later, this can wait.

In this case also createNewNoTtsAvailableDialog() should be removed.

TheLastProject:
>If you believe that you will look at this later, this can wait.

I wish I could, but as I described I can no longer test this issue as since today's OS update there is a TTS engine built into /e/OS which I can't disable or remove.

>In this case also createNewNoTtsAvailableDialog() should be removed.

Good catch, thanks

gerhardol (approved):

Issue #352, opened 2015-10-30
Cannot edit workouts


embeh:
Despite the GUI label "Manage workouts \ Download/edit/remove workouts" in Settings, there seems to be no way to modify an existing workout. All you can do is delete what you have and start over, which is a pain if you have created a complex workout.

RunningCl0ud:
Therese is no problem with my SG S5? What kind of Android do you use?

jonasoreland:
existing workouts can be modified directly on the start->advanced screen,
there you can (atleast) modify e.g repeat count, phase, duration etc

have you tried that ?

--

But I agree, it would be nice if you could get to edit (rather than just
create) a workout.

/Jonas

embeh:
@RunningCl0ud This is on Android 4.4

@jonasoreland Yes, I can modify the numbers but cannot add or remove a step

RunningCl0ud:
@embeh Ok I see what you mean. It could be an extra feature to "edit" f.ex put in an extra step somewhere in the existing workout. That does not seem to be doable for the moment and I am on Android 5.0. 
If you have Garmin Connect you can easily edit the workout there. But I see your point and it is a good one I think. 
And if you just want to modify existing steps @jonasoreland s idea should work?

embeh:
@RunningCl0ud Exactly. I recently had two cases where I created a workout with maybe 15 steps (more than could fit on a single screen, with several repetitions), and each time the app would insert an additional step somewhere above - which I did not notice until afterwards. All I could do then was to set these back to 0 minutes. But the voice coaching would still deal with those steps, which was all somewhat confusing. I realize I can edit the json file but that is not really what you want an app user to do, right?

davidedelvento:
I think this whole way of editing workouts should be refreshed. See #567 

@davidedelvento

gerhardol:
This is handled in #587 
Closing this issue

davidedelvento:
@gerhardol I think #587 is totally different: that is not for long-term editing but for throw-away changes done while running.
This is more for things like #567 #583 and similar things I'm working on now. No need to reopen, but just making it clear we're talking about different things.

anusser:
First, thanks a lot for this great app!

As there hasn't been any activity in this issue since a few years, I just wanted to bump it. For me this issue is still there and the only drawback in an until now flawless app.

gerhardol:
It is possible to change the number of steps and types for a workout. 
Is it really that important to add or remove new steps to a workout?

[...]

PR #984, opened 2020-11-15
Added edit button for existing advanced workouts

Upabjojr:
This PR adds the possibility of opening the existing advanced workouts in edit mode.

Fixes https://github.com/jonasoreland/runnerup/issues/352

### DIFF
--- a/app/res/layout/manage_workouts.xml
+++ b/app/res/layout/manage_workouts.xml
@@ -62,6 +62,17 @@
           android:text="@string/Share"
           android:textSize="16sp"
       />
+
+      <Button
+          android:id="@+id/edit_workout_button"
+          style="@style/ButtonText"
+          android:background="@drawable/btn_blue"
+          android:layout_width="0dp"
+          android:layout_weight="1"
+          android:layout_height="match_parent"
+          android:text="@string/Edit"
+          android:textSize="16sp"
+      />
     </TableRow>
 
     <TableRow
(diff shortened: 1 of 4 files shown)


gerhardol (approved):
Looks good
Can you rebase on master?

Upabjojr:
OK, I think it's rebased.

gerhardol:
Thanks!

opoto:
Hi. Nice PR, thanks!
However I think the "Discard" button in the edit screen is confusing: I was assuming it would discard edits, while it actually deleted the workout. Is this intentional? I think deletion should not be possible from edit screen, but only from the workout list, where there is a proper "Delete" button.

Upabjojr:
Well, all this PR does is adding an edit button to allow the creation screen to reappear even after the workout has been created. The edit screen is exactly the same as the creation screen.

Maybe in this case the discard button should be hidden, indeed.

opoto:
Better than removing that button, I would expect it to be present and implement what was expected: cancel current edits and restore previous workout state. For a new workout it means cancelling workout creation, but for an existing edited one it would mean restore state as before entering the edit screen. 
I understand this might be more complex to implement.

Upabjojr:
> Better than removing that button, I would expect it to be present and implement what was expected: cancel current edits and restore previous workout state. For a new workout it means cancelling workout creation, but for an existing edited one it would mean restore state as before entering the edit screen.
> I understand this might be more complex to implement.

The problem is that it is in autosave mode.

opoto:
Ok, I understand it would imply deeper changes.
Removing the button is the best option for now then.
Thanks

PR #986, opened 2020-11-15
Add preference for custom MapBox API key for F-Droid users

Upabjojr:
Adding a text-preference for custom MapBox API keys.

This allows F-Droid users to use the mapbox widget by adding a custom API key.

### DIFF
--- a/app/res/values/pref_keys.xml
+++ b/app/res/values/pref_keys.xml
@@ -112,6 +112,7 @@
     <string name="pref_path_simplification_algorithm">pref_path_simplification_algorithm</string>
 
     <string name="pref_mapbox_default_style">pref_mapbox_default_style2</string>
+    <string name="pref_mapbox_custom_api_key">pref_mapbox_custom_api_key</string>
     <string name="pref_runneruplive_active">pref_runneruplive_active</string>
     <string name="pref_runneruplive_serveradress">pref_runneruplive_serveradress</string>
 
--- a/app/res/xml/settings.xml
+++ b/app/res/xml/settings.xml
@@ -373,6 +373,15 @@
             android:minHeight="48dp"
             android:persistent="true"
             android:title="@string/Mapbox_default_style" />
+
+        <org.runnerup.widget.TextPreference
+            android:defaultValue=""
+            android:inputType="text"
+            android:key="@string/pref_mapbox_custom_api_key"
+            android:minHeight="48dp"
+            android:persistent="true"
+            android:title="@string/custom_mapbox_api_key" />
+
     </PreferenceScreen>
 
     <PreferenceScreen
--- a/app/src/main/org/runnerup/view/DetailActivity.java
+++ b/app/src/main/org/runnerup/view/DetailActivity.java
@@ -21,11 +21,13 @@ import android.annotation.SuppressLint;
 import android.content.ContentValues;
 import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.database.Cursor;
 import android.database.sqlite.SQLiteDatabase;
 import android.graphics.Color;
 import android.net.Uri;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 import android.text.TextUtils;
 import android.util.Log;
 import android.view.LayoutInflater;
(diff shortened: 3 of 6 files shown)


gerhardol:
One reason to remove MapBox libs from F-Droid was that it is unsure if MapBox meets the F-Droid requirements.
(Especially after they have changed the license, not updated in RunnerUp though.)
I saw some discussion in F_Droid about that and the binaries were not used anyway.

Do you have any more info?

Upabjojr:
Didn't realize MapBox is close-sourcing its API. That's bad news.

Maybe it's better to see if there is a way to open the track in OsmAnd?

gerhardol:
MapBox is closing some parts, a quick search for how to exclude those parts:
https://forum.f-droid.org/t/updated-distribution-license-for-mapbox-mobile-maps-sdk/10201/6

I would like to have a solution with a map in the Play version still, but MapBox is not a requirement and the Play/F-Droid solutions could be separate.

Uatschitchun:
https://github.com/jonasoreland/runnerup/issues/1052

gerhardol:
Considering API keys is required also when building MapBox, I doubt this will be possible in the free version.

Upabjojr:
> Considering API keys is required also when building MapBox, I doubt this will be possible in the free version.

Aren't there any free map widgets available?

(a bit out of topic, maybe some better interaction with OsmAnd would be nice to have)

gerhardol:
> > Considering API keys is required also when building MapBox, I doubt this will be possible in the free version.
> 
> Aren't there any free map widgets available?

Probably for the viewer, but you need the map data too.
Did not find anything suitable a few years ago.

> (a bit out of topic, maybe some better interaction with OsmAnd would be nice to have)

Share gpx should be possible if OsmAnd implements it. (Maybe there is an intent already, have not checked the inner workings.)

Upabjojr:
> Probably for the viewer, but you need the map data too.
> Did not find anything suitable a few years ago.

This package is on F-Droid:
https://f-droid.org/en/packages/de.storchp.opentracks.osmplugin/
They look like they have maps integrated with tracks. Did not inspect their source code so far.

Upabjojr:
> Share gpx should be possible if OsmAnd implements it. (Maybe there is an intent already, have not checked the inner workings.)

Yes, I have to save the GPX to the storage space and then import it into OsmAnd. Maybe RunnerUp could be added an option to save the GPX track to the storage space directly?

gerhardol:
> > Share gpx should be possible if OsmAnd implements it. (Maybe there is an intent already, have not checked the inner workings.)
> 
> Yes, I have to save the GPX to the storage space and then import it into OsmAnd. Maybe RunnerUp could be added an option to save the GPX track to the storage space directly?

You can use the file synchronizer to automatically export to gpx/tcx.

PanderMusubi:
So there are three cases:
1) for the normal build, use the API key that is shipped
2) for the F-Droid build, do not use the shipped API key
3) use a personal API key (overriding the cases above for any build)

Upabjojr:
What about using osmdroid for the F-Droid version?

gerhardol:
Thanks, but will close this PR
The FDroid version supports OSMdroid, so there is an option there
Of course, you may build your own 'free' package with a private key...

Upabjojr:
> Thanks, but will close this PR The FDroid version supports OSMdroid, so there is an option there Of course, you may build your own 'free' package with a private key...

Well, I am the author of the PR introducing osmdroid. Forgot to close this PR.

Upabjojr:
By the way, the last F-droid version was released more than 1 year ago and it doesn't include osmdroid yet. What about a new release? @gerhardol

gerhardol:
> By the way, the last F-droid version was released more than 1 year ago and it doesn't include osmdroid yet. What about a new release? @gerhardol

That is what I am working on now. As usuual, most of the work is to adapt to new Google requirements to be able to update in Play. There were some other PRs that requires some job too. The work required to do something quick has built up faster than I have been able to fix it.
It would be better if the maintainer was actually using the app...

Will hopefully make a beta soon, that will require some translation changes though.
But FDroid maintainers changed to release on Git tag, so I guess FDroid will be published after the beta....

PR #989, opened 2020-11-21
Hide 'Discard' Button when editing existing advanced workouts

Upabjojr:
Responding to the idea of comment https://github.com/jonasoreland/runnerup/pull/984#issuecomment-731596314


gerhardol (commented):
The Save button should be expanded, similar to other buttons

image

Upabjojr:
> The Save button should be expanded, similar to other buttons

How do you do that?

gerhardol:
> How do you do that?

Magic
Similar to the share/edit row before
INVISIBLE -> GONE addresses that

gerhardol (approved):
Will squash merge, unless manually squashed
Thanks!

PR #1003, opened 2021-02-03
Adding a Hebrew translation

tzagim:


gerhardol:
Thanks
The sources for translations are stored in Transifex, but as that was just added (requested by you?) I can push it, I believe

tzagim:
Yes it's me, can you define me as a translator?

gerhardol:
>Yes it's me, can you define me as a translator?

You  are invited as a reviewer for Hebrew too

Uploaded to Transifex, pulled the changes down. Will merge these changes in https://github.com/jonasoreland/runnerup/pull/1004
Note that there were an error in the strings.xml (easily fixed) and that Transifex did not accept plurals (4 requested.
If you add that to Transifex I will update

tzagim:
I fixed it.
This is a known problem in Hebrew, not written a digit in the singular or in two

gerhardol:
updated from Transifex, will merge in my PR (you are the author for the commit)

Is the short name 'he' in Android, I have seen that 'iw' should be used?

If there are right-to-left issues, I hope you can address those...

gerhardol:
Will rename the languagelocally to iw, lnt warning

```

C:\dev\gc\runnerup\app\res\values-he: Warning: The locale folder "he" should be called "iw" instead; see the java.util.Locale documentation [LocaleFolder]

   Explanation for issues of type "LocaleFolder":
   From the java.util.Locale documentation:
   "Note that Java uses several deprecated two-letter codes. The Hebrew ("he")
   language code is rewritten as "iw", Indonesian ("id") as "in", and Yiddish
   ("yi") as "ji". This rewriting happens even if you construct your own
   Locale object, not just for instances returned by the various lookup
   methods.

   Because of this, if you add your localized resources in for example
   values-he they will not be used, since the system will look for values-iw
   instead.

   To work around this, place your resources in a values folder using the
   deprecated language code instead.

   https://developer.android.com/reference/java/util/Locale.html

```

gerhardol:
included in #1004

PR #1012, opened 2021-03-14
Ensure HR value is 0 when HR is disconnected.

opoto:
HR value is used in HRMStateChangeFeedback to choose between connection_lost and conection_restored. On my mobile with BLE HR band, the HR value after disconnection is the latest collected value, which leads to conection_restored message on both disconnection and reconnection events. This change should fix this issue.

### DIFF
--- a/app/src/main/org/runnerup/tracker/Tracker.java
+++ b/app/src/main/org/runnerup/tracker/Tracker.java
@@ -892,7 +892,7 @@ public class Tracker extends android.app.Service implements
         }
 
         HRProvider hrProvider = trackerHRM.getHrProvider();
-        if (hrProvider == null)
+        if ((hrProvider == null) || !hrProvider.isConnected())
             return null;
 
         // now is elapsed nanosec, no sensor adjust needed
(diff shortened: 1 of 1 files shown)


gerhardol:
The same change should be applied to getCurrentHRValue()

However, if not connected, the return value should probably be null?

opoto:
You're correct on both points! Updating...

opoto:
now returning null, which will ultimately be converted to 0 in Workout.getHeartRate()

app/src/main/org/runnerup/tracker/Tracker.java
```diff
@@ -899,7 +899,7 @@ private Integer getCurrentHRValueElapsed(long now, long maxAge) {
         if (now > hrProvider.getHRValueElapsedRealtime() + maxAge * NANO_IN_MILLI)
```
gerhardol, line 899: I expected the isConnected here...

Change that and I will merge

opoto:
Makes even more sense when testing hrProvider presence (lines 895 & 907).
Thanks

gerhardol (approved):

opoto:
Hi. Is anything more expected from this PR to be merged? Thanks

gerhardol:
> Hi. Is anything more expected from this PR to be merged? Thanks

Sorry, I was busy.
Thanks!

PR #1017, opened 2021-04-11
Fixes #1016 - always insert Authorization header in requests

### LINKED ISSUE #1016: WebDAV synchronizer fails to authenticate with valid credentials
I've been trying to get the WebDAV to authenticate properly with my WebDAV server.

From my investiation into the issue it seems like OkHttp make one request without the `Authorization` header and then it's supposed to add the `Authorization` header on HTTP 401, however this those not seem to happen.
I have not been able to find the cause of this, but it seems like a better flow in this instance would be to always provide the credentials for the synchronizer.

I'll shortly make a PR to implement this behavior which avoids having to make 2 requests for every action, which also solves the issue authentication issue I've been facing.

BlackVoid:
This PR fixes #1016 and ensures that we always send the Authorization header on every request, the default authenticator always returns null, so if authentication fails it will not make additional retries and as such preventing an infinite loop.

### DIFF
--- a/app/src/main/org/runnerup/export/WebDavSynchronizer.java
+++ b/app/src/main/org/runnerup/export/WebDavSynchronizer.java
@@ -20,6 +20,7 @@ package org.runnerup.export;
 import android.content.ContentValues;
 import android.database.Cursor;
 import android.database.sqlite.SQLiteDatabase;
+import android.text.TextUtils;
 import android.util.Log;
 
 import androidx.annotation.ColorRes;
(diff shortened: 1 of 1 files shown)


app/src/main/org/runnerup/export/WebDavSynchronizer.java
```diff
@@ -189,12 +189,9 @@ public Status connect() {
...
     private OkHttpClient getAuthClient() {
-        return new OkHttpClient().newBuilder().authenticator((route, response) -> {
-            if (response.request().header("Authorization") != null) {
-                return null; // Give up, we've already failed to authenticate.
-            }
-            String credential = Credentials.basic(username, password);
-            return response.request().newBuilder().header("Authorization", credential).build();
+        return new OkHttpClient().newBuilder().addInterceptor(chain -> {
+            String credentials = Credentials.basic(username, password);
+            return chain.proceed(chain.request().newBuilder().header("Authorization", credentials).build());
```
gerhardol, line 194: Are there WebDAV without authorization that will be affected by this?
Check if creadentials are set bwfore adding?

BlackVoid: Yeah, I guess that use case would be affected, I added a check if either username or password has been set.

gerhardol (commented):
Seem to work for me, I had some problems with the WebDAV I am testing with.
It also seems like user/pass is required.
Without a deeper investigation on OkHttp, if the previous implementation was supposed to be aborted after two attempts: Is there a risk for infinitive retries?

BlackVoid:
> Without a deeper investigation on OkHttp, if the previous implementation was supposed to be aborted after two attempts: Is there a risk for infinitive retries?

The default authenticator is ["None" authenticator](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-authenticator/-n-o-n-e/). So it will not make any additional attempts when getting a 401 and that's also the behavior I saw when testing with invalid credentials. So I'm pretty certain that this will not introduce any infinite loops.

gerhardol (approved):

PR #1019, opened 2021-04-21
Add current battery level to HR connection status

matgoebl:
The battery level of HR sensors is only shown in a popup (when it is low) and while setting up the HR device.
Having it in the connection status line (displayed before starting the run) helps keeping track of remaining battery life.

### DIFF
--- a/app/src/main/org/runnerup/view/StartActivity.java
+++ b/app/src/main/org/runnerup/view/StartActivity.java
@@ -1009,6 +1009,8 @@ public class StartActivity extends AppCompatActivity
             Integer hrVal = mTracker.getCurrentHRValue();
             if (hrVal != null)
                 str.append(" ").append(hrVal);
+                Integer batteryLevel = mTracker.getCurrentBatteryLevel();
+                str.append(" ").append(batteryLevel).append("%");
         }
         return str.toString();
     }


gerhardol:
seem OK

app/src/main/org/runnerup/view/StartActivity.java
```diff
@@ -1009,6 +1009,8 @@ private String getHRDetailString() {
             Integer hrVal = mTracker.getCurrentHRValue();
             if (hrVal != null)
                 str.append(" ").append(hrVal);
+                Integer batteryLevel = mTracker.getCurrentBatteryLevel();
+                str.append(" ").append(batteryLevel).append("%");
```
gerhardol, line 1013:
```suggestion
            if (hrVal != null) {
                str.append(" ").append(hrVal);
                Integer batteryLevel = mTracker.getCurrentBatteryLevel();
                str.append(" ").append(batteryLevel).append("%");
            }

```

gerhardol (changes requested):
It is bad to not have braces, but this is how the original style was...

matgoebl:
> It is bad to not have braces, but this is how the original style was...

Thanks for being so watchful :+1:

gerhardol (approved):

PR #1025, opened 2021-05-24
Resolve disallowed_agent error in Strava login with Google account

### LINKED ISSUE #791: Strava, Runalyze: Runkeeper login with Google: disallowed_useragent
Login to Strava, Runalyze or RunKeeper using Google fails with disallowed_useragent
It is possible to login to the services using direct login or Facebook.

Note: This will be seen when updating to 2.0.0.5 for Strava, as the authentication is changed (Strava will drop the old auth in October). So this is not a problem with 2.0.0.5

Strava will require a redesign, to auth in a different way. Strava has some hints for this.
For Runalyze, I am not aware of how this can be done, direct login recommended.
RunKeeper has disallowed the open API and will likely shut it down anyway so nothing should be done. (To be removed in #701)

Rejna:
Should fix #791. I haven't checked the integration with services other than Strava.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,1 +1,1 @@
-Please see the [wiki](https://github.com/jonasoreland/runnerup/wiki)
+Please see the [wiki](https://github.com/jonasoreland/runnerup/wiki/CONTRIBUTING)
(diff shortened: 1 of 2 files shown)


gerhardol:
Seem to work. Tried RunAlyze and Dropbox. Cannot see a reason the updated user agent will break the other services.
Not tried Facebook etc.

I had got the impression that the embedded browser had to be replaced for Google login to work.

Thanks!

gerhardol (approved):

Issue #991, opened 2020-11-25
Heart rate zone target unexpected behavior


Chushuhuch:
I'm creating custom workout with heart rate zone target. Each time I choose some target zone and press ok it is saved and shown correctly in the interface.
But after I enter editing the step again, the zone is different or not set at all.
Also it seems that during workout the zone was also wrong since I didn't get any message from the app that my pulse is too high. At least it worked for the pulse too low. So maybe this is broken as well.

gerhardol:
It is only possible to edit the lower limit, the upper is set from previous lower.

The UI should not allow chaning the upper limit.

Chushuhuch:
I'm not sure. Are you talking about HR zone limits set up?
I've meant workout creation feature. I believe setting HR zone limits is a different thing.
Do you mean that HR targets are lower bound only?

gerhardol:
HR zones are created using lower zones only so that may be something different.

I can edit HR zones, not tried the limits.

Chushuhuch:
Oh, indeed. I can edit the upper limit, but the change is not saved.
Still, at least in the UI any zone is a range. And in general if I'm setting the goal to run in zone 3, my personal goal is usually not to go to zone 4 accidentally. And probably for zones 4 and 5 it is opposite - not to go too low. So I personally would expect alerts both for having too high and too low HR.

Also just found that saved HR zone is +2 to the selected. If I choose zone 1, after closing and reopening it becomes zone 3. And zones 4 and 5 become empty.

gerhardol:
Closed by #991

PR #1028, opened 2021-07-27
Fix bugs #991 in HRZones

sv2ramanan:
These changes address bug/issue #991 . 

Part 1 addresses "The UI should not allow changing the upper limit."

1. HRZonesActivity.java: 
  Make the high range of HR Zones uneditable.
   Map the low end of a zone to the high end of the previous zone on-the-fly
   A first attempt to make sure the zones are ordered (increasing)

Part 2  fixes a bug in zones: "saved HR zone is +2 to the selected. If I choose zone 1, after closing and reopening it becomes zone 3. And zones 4 and 5 become empty.". This behavior seems to derive from different zero- versus one-based numbering: 1 for the zones (for users) and 0 programmatically. This results in changes of 2 in HR zones when editing zones as targets of intervals.

2. StepButton.java:
   When HR is the target in an interval, the zone changes by +2 each time it is re-edited.
   This is fixed by subtracting 2 from the results of the match of hrZones from its adapter
   This is probably not the best way to write the solution, but it prevents app crashes as well.

### DIFF
--- a/app/src/main/org/runnerup/view/StepButton.java
+++ b/app/src/main/org/runnerup/view/StepButton.java
@@ -300,10 +300,11 @@ public class StepButton extends LinearLayout {
                         targetHrz.setEnabled(true);
                         targetHrz.setVisibility(View.VISIBLE);
                         if (target != null) {
-                            int matchedZone=hrZonesAdapter.hrZones.match(target.minValue,
-                                    target.maxValue)-2;
-                            if (matchedZone<0)
-                                matchedZone=0;
+                            int matchedZone = hrZonesAdapter.hrZones.match(target.minValue,
+                                    target.maxValue) - 2;
+                            if (matchedZone < 0) {
+                                matchedZone = 0;
+                            }
                             targetHrz.setValue(matchedZone);
                         } else {
                             targetHrz.setValue(0);
@@ -311,8 +312,7 @@ public class StepButton extends LinearLayout {
                         break;
                     default:
                         targetPaceLo.setEnabled(false);
-                        targetPaceHi
-                                .setEnabled(false);
+                        targetPaceHi.setEnabled(false);
                         targetHrz.setEnabled(false);
                         break;
                 }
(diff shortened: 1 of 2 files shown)


app/src/main/org/runnerup/view/StepButton.java
```diff
@@ -300,8 +300,11 @@ public int preSetValue(int newValue) throws IllegalArgumentException {
                         targetHrz.setEnabled(true);
                         targetHrz.setVisibility(View.VISIBLE);
                         if (target != null) {
-                            targetHrz.setValue(hrZonesAdapter.hrZones.match(target.minValue,
-                                    target.maxValue));
+                            int matchedZone=hrZonesAdapter.hrZones.match(target.minValue,
+                                    target.maxValue)-2;
+                            if (matchedZone<0)
```
gerhardol, line 305: The code base is inconsistent, but I want braces at all if and spaces around operators

gerhardol (commented):
Seem OK, have briefly played around with it

How to handle upper for zone 5?
It is probably OK to lock it to MHR
It is possible to set lower for zone 5 to > than MHR though... (just increase MHR?)

sv2ramanan:
1. Fixed braces and spaces in code
2. Set low HR for zone 5 to be maxHR-1 at most
3. Set low HR for any zone to be maxHR - 1 - (noZones-zone), so each Zone has 1 HR beat range at least

app/src/main/org/runnerup/view/StepButton.java
```diff
@@ -300,16 +300,19 @@ public int preSetValue(int newValue) throws IllegalArgumentException {
                         targetHrz.setEnabled(true);
                         targetHrz.setVisibility(View.VISIBLE);
                         if (target != null) {
-                            targetHrz.setValue(hrZonesAdapter.hrZones.match(target.minValue,
-                                    target.maxValue));
+                            int matchedZone = hrZonesAdapter.hrZones.match(target.minValue,
```
gerhardol, line 303: Please add a comment here

gerhardol (approved):
It is still possible to set zones in wrong order, but it is better than before
Do you want to try fix it or merge as is?

After adding the comment, do you want me to squash merge or clean up the commits yourself first?

PR #1033, opened 2021-08-12
Record activity to blockchain.

winder:
New "Algorand" account type which allows recording activity data to a blockchain.

I really enjoyed working on the RunnerUp code, the design and architecture to be very approachable. This was also the first time I've worked with an Android application, figuring out things like creating the view XML file and working with the emulator was the most challenging part.

I'm going to continue testing, but I think everything is currently working. One thing I haven't figured out yet is how to upgrade my play store installation with an APK I built. Can this be done?

I'm using a wallet application to manage authentication, so there is really no configuration. I made a simple "enable" auth so that users didn't have it turned on when they didn't want/expect it. During the upload it brings you to another app to write the data.
Screenshot_20210812-132714


app/src/main/org/runnerup/export/SyncManager.java
```diff
@@ -259,6 +275,11 @@ public void connect(final Callback callback, final String name) {
                 }, synchronizer, s.authMethod);
                 return;
 
+            case ERROR:
+                if (s.ex != null) {
+                    // TODO: Display error message if there is an atta
```
gerhardol, line 280: ?

app/src/main/org/runnerup/export/format/Summary.java
```diff
@@ -0,0 +1,81 @@
...
+            if (c.moveToFirst()) {
+                Long startTime = c.getLong(1);
+                Long distance = c.getLong(2);
+                Long durationTime = c.getLong(3);
+                Sport sport = Sport.valueOf(c.getInt(4));
+
+                Date t = new Date(startTime * 1000);
+                String startStr = simpleDateFormat.format(t);
+                String distStr = formatter.format(Formatter.Format.TXT, Dimension.DISTANCE, distance);
+                distStr += " " + formatter.getUnitString();
```
gerhardol, line 68: if unit is not included in the formatter, it should be added

app/test/java/org/runnerup/export/format/SummaryTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @After
+    public void teardown() {
+        db.close();
+    }
+
+    @Test
+    public void TestSummary() throws IOException {
+        String str = sum.Summary(5);
+        assertEquals("RUNNING: 2.29mi at Sun, 8 Aug 2021 05:02 PM EDT for 25m8s. Pace: 11:00/mi", str);
```
gerhardol, line 59: Is the locale forced here?
It should be independent of the settings.

app/src/main/org/runnerup/export/AlgorandSynchronizer.java
```diff
@@ -0,0 +1,212 @@
...
+            return s;
+        }
+    }
+
+    /**
+     * check if algorand wallet is installed.
+     * @return
+     */
+    public Boolean walletReady() {
+        // TODO: How to test if the Algorand Wallet is installed
```
gerhardol, line 147: Something more here would be nice, to at least explain what Algorand Wallet is

gerhardol (commented):
Very briefly looked at the code, looks fine
Not much that is synced though...

app/src/main/org/runnerup/export/SyncManager.java
winder: I'm attaching an exception to the status that says something like "Algorand Wallet applicaiton is not installed.", but we don't currently check the `ERROR` case. I'm thinking a popup here would be good for telling the user if something went wrong. Maybe other accounts could have errors like this if the OAuth token was revoked.

app/src/main/org/runnerup/export/AlgorandSynchronizer.java
winder: Yes, I'd like to figure out a better way for this to validate the configuration. I noticed that the FileSynchronizer kludged it a little bit, and I needed to repeat that kludge. One thing I didn't notice until the end of my integration was the way the FileSynchronizer checked for it's required permissions:
```
            case FILEPERMISSION:
                checkStoragePermissions(mActivity);
                askFileUrl(l);
```

I'm wondering if the interface could do with two more methods:
`validatePermissions(Context c)` for post-config permission validation
`updateAuthConfig(ContentValue cv)` to replace the hard coded `FileSynchronizer.contentValuesToAuthConfig(tmp)` type calls.

gerhardol: probably OK

winder:
Went for a run and.... it works!

https://algoexplorer.io/tx/NB7ZA3BFNCSZPYOADBWTXAAGHSW6XLFHIBG6IDD3X2BKYWXZ52TQ

gerhardol:
> Went for a run and.... it works!

I can use it locally too (in emulator). What about the fee at purestake.io?
(and I would much prefer that there is a message if Algorand Wallet is not installed)

gerhardol:
> One thing I haven't figured out yet is how to upgrade my play store installation with an APK I built. Can this be done?

No, the signing key is private.
The debug build has a separate id though, so you can have them installed in parallel.
It is possible to setup private groups in Play but you still need the key in the first place...

winder:
> (and I would much prefer that there is a message if Algorand Wallet is not installed)



> > Went for a run and.... it works!
> 
> I can use it locally too (in emulator). What about the fee at purestake.io?
> (and I would much prefer that there is a message if Algorand Wallet is not installed)

There a fee to create a transaction, on this blockchain it is currently about 0.1 cents (as in less than 1 cent, not 10 cents).

If you (or anyone) share an algorand account here, I can send you 25 cents worth of algos which should cover you for hundreds of activities. You can create the account in the Algorand Wallet app.

winder:
@gerhardol I updated the Synchronization interface and improved the error handling. Please let me know what you think.

app/src/main/org/runnerup/export/AlgorandSynchronizer.java
```diff
@@ -139,26 +139,22 @@ public Status connect() {
...
+        if (!appInstalledOrNot("com.algorand.android")) {
+            throw new Exception(mContext.getString(R.string.algorand_app_required));
+        }
+    }
 
+    private boolean appInstalledOrNot(String uri) {
+        PackageManager pm = mContext.getPackageManager();
         try {
-            mContext.startActivity(intent);
+            pm.getPackageInfo(uri, PackageManager.GET_ACTIVITIES);
```
opoto, line 152: I don't think this will work when RU will be upgraded to Android target api 30:
https://developer.android.com/training/package-visibility

winder: Thanks, I had no idea... I replaced this with a call to invoke the activity and handle the `ActivityNotFound` exception. This should not require any extra permissions but requires more user interaction. Specifically if the activity is available runnerup will switch to the other app, and the user must switch back.

opoto: Looks better, thanks!

gerhardol (approved):
I have had a hard time testing this but all seem fine, just nit pick

Do you want to squash before merge or should I squash to one commit?

app/src/main/org/runnerup/export/AlgorandSynchronizer.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Override
+    public void validatePermissions(ValidateCallback callback, AppCompatActivity activity, Context context) {
+        new AlertDialog.Builder(mContext)
+                .setTitle(mContext.getString(R.string.test_configuration))
+                .setMessage(mContext.getString(R.string.algorand_app_test))
+                .setPositiveButton(R.string.OK, (dialog, which) -> {
+                    // Send transaction to Algorand Wallet application.
+                    String query = scheme + "?amount=0&note=If you see this text the test is a success, return to RunnerUp!";
```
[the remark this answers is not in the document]
winder: I'm not sure what you mean by preview will fail? I can add the localization.

winder:
Thanks for the feedback! I've been testing for a couple weeks now and day-to-day it works fine. There is one issue for the initial import that I should probably be addressed before this gets merged. When a whole batch of updates need to be made the user needs to approve/send each transaction, when the apps switch it only pops up the first (or first and last if I'm quick).

I'm guessing this is a property of the Algorand Wallet, maybe even a bug. I'm hoping to find a way to send these serially to workaround the problem.

gerhardol:
Any updates?

gerhardol:
Time to finalize?

winder:
Really sorry to have wasted your time, but I don't see myself getting motivated to get this into a state where it's fully working.

I'm also not sure the interface I chose is going to be stable without revisiting it periodically.

I'll create a new PR if I get to it again.

gerhardol:
No problem.

PR #1036, opened 2021-08-21
Hr zones ubound fix; 

sv2ramanan:
This is the same branch from before to fix HR zones, but now with changes to the code so that hr zones will not go out of order. I have also added a comment where it was recommended.

However, I have somehow got the github branches wrong, and got totally muddled (or something). So I am submitting a new pull request. Apologies for the trouble. Also, please go ahead with the squash commit, as I may not get that right too.


gerhardol:
Thanks!

PR #1048, opened 2021-10-03
Clean up preferences, add categories & summaries for declaration and …

Uatschitchun:
This PR cleans and beautifies up the preferences with adding them into categories and add summaries for declaration and explanation.

PR #1050, opened 2021-10-04
Final: Preferences updates

Uatschitchun:
Second try for the preference updates. Suggestions from first try https://github.com/jonasoreland/runnerup/pull/1048 added

### DIFF
--- a/app/assets/about.html
+++ b/app/assets/about.html
@@ -8,6 +8,7 @@
 <p><a href="https://github.com/jonasoreland/runnerup/wiki">RunnerUp wiki</a></p>
 <p><a href="https://github.com/jonasoreland/runnerup/issues">RunnerUp issue list</a></p>
 <p><a href="https://rawgit.com/jonasoreland/runnerup/master/app/assets/changes.html">Release History</a></p>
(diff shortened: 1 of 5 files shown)


Uatschitchun:
I would suggest to "risk" missing translations, as this will hopefully motivates translators to update. German translation had a huge update in transifex and helps understanding settings and usage together with this preference update. 
Translating the new strings in german is a matter of minutes
What about adding a link to https://www.transifex.com/runner-up/runner-up-android/dashboard/ into ABOUT?

gerhardol (approved):
I will have to review the pending changes before merging though.
Maybe the current beta can be promoted to production soon (the previous beta had a crash.)

Uatschitchun:
@gerhardol am going for a run this morning and will test the latest fix regarding cadence.

Uatschitchun:
Will this not be part of 2.4.3.1?

gerhardol:
Finally got to cleanup this PR.
There were many minor issues that was difficult to comment on, so I pushed changes
There will be a few other changes in some private PRs too

The big issue is that the SettingsActivity is deprecated and need to be replaced with fragments,

PR #1053, opened 2021-10-04
Implement a catch to back button for to prevent app from quitting dir…

Uatschitchun:
This one is an example on how to catch 'one too many' back button presses always quitting RU. (happens to me all the time ;)

It would be best suited somewhere in main activity, I guess, but I don't know the right place or file ;)

### DIFF
--- a/app/src/main/org/runnerup/view/SettingsActivity.java
+++ b/app/src/main/org/runnerup/view/SettingsActivity.java
@@ -23,11 +23,13 @@ import android.content.SharedPreferences;
 import android.content.res.Resources;
 import android.os.Build;
 import android.os.Bundle;
+import android.os.Handler;
 import android.preference.CheckBoxPreference;
 import android.preference.Preference;
 import android.preference.Preference.OnPreferenceClickListener;
 import android.preference.PreferenceActivity;
 import android.preference.PreferenceManager;
+import android.widget.Toast;
 
 import org.runnerup.BuildConfig;
 import org.runnerup.R;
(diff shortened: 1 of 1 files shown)


app/src/main/org/runnerup/view/SettingsActivity.java
```diff
@@ -128,4 +130,24 @@ public static boolean hasHR(Context ctx) {
...
         return false;
     };
+
+    private Boolean exit = false;
+    @Override
+    public void onBackPressed() {
+        if (exit) {
+            finish(); // finish activity
+        } else {
+            Toast.makeText(this, "Press Back again to Exit.",
```
gerhardol, line 140: Should  be translatable

app/src/main/org/runnerup/view/SettingsActivity.java
```diff
@@ -128,4 +130,24 @@ public static boolean hasHR(Context ctx) {
...
+            new Handler().postDelayed(new Runnable() {
+                @Override
+                public void run() {
+                    exit = false;
+                }
+            }, 3 * 1000);
+
+        }
+
+    };
```
gerhardol, line 152: nit pick: remove ; and empty spaces

gerhardol (approved):
SettingsActivity will need to be rewritten with fragments, should also be a normal activity so this is not needed.
But exiting has happened to me too...

Some minor comments to fix, you may squash.

Uatschitchun:
What I would ideally like is back button always reverting to Start/Activity Screen and only exit from there with 2 times back button.

If only start screen gets this catch (2 times back for quitting) and all other screens jump back to start screen instead of quitting.

gerhardol:
> What I would ideally like is back button always reverting to Start/Activity Screen and only exit from there with 2 times back button.
> 
> If only start screen gets this catch (2 times back for quitting) and all other screens jump back to start screen instead of quitting.

That would be the case if settings was a separate activity.

I would like to merge basic,interval workout to advanced and remove feed (not much use any more), any suggestions for updating the UI?

PR #1055, opened 2021-10-05
Fix distancepicker

Uatschitchun:
Setting distances was limited to 59 unitMeters.
There are sports that do more than 100km. So I set digits to 3 and maxValue to 999.

Together with this I speed up the long press. Was way too slow


gerhardol (approved):
Note the lame handling of fractions, this is really 0-1609m if miles are used.
(I tried fixed that at one point of time, but what unit to use for fraction of a mile? probably yards)

Uatschitchun:
A mile exactly is 1,609.344 metres and 1760 yards. I don't think those fractions are needed, as running on a sports track for instance is a ¼ mile.
I'm going to ask in r/running :-)

PR #1056, opened 2021-10-05
Wip: Fix HR in DB. Max & avg were swapped

Uatschitchun:
Stumbled upon the fact that max_hr and avg_hr were swapped while looking at the DB's activity

### DIFF
--- a/common/src/main/java/org/runnerup/common/util/Constants.java
+++ b/common/src/main/java/org/runnerup/common/util/Constants.java
@@ -41,8 +41,8 @@ public interface Constants {
             String NAME = "name";
             String COMMENT = "comment";
             String SPORT = "type";
-            String MAX_HR = "avg_hr";
-            String AVG_HR = "max_hr";
+            String MAX_HR = "max_hr";
+            String AVG_HR = "avg_hr";
             String AVG_CADENCE = "avg_cadence";
             String META_DATA = "meta_data";
             String DELETED = "deleted";


gerhardol (changes requested):
Regardless of what the fields are called, max hr is read/write from MAX_HR.
If this is swapped, the max/avg will be swapped in existing workouts.
This would require a migration of the database.

Add a comment instead.

Uatschitchun:
Max & avg HR are swamped in DB. My test recording yesterday to test cadence resulted in 183 avg HR & 152 max HR for that activity.

Yes, for to fix existing workouts this would need a migration in DB. But should RU still record swapped measures into DB?

I fear that every export (#1054) uses those swapped values...

I see 3 possibilities:
1.) just swap and record new workouts correctly from now on, leaving the existing workouts in a false state
2.) swap for new workouts to be recorded correctly and do a DB migration initiated in the "what's new" update thread.
* if I see correctly, there's no DB scheme version in the DB. So no way to update to a new scheme version or testing for version numbers
3.) Leave it as it is and record false values further on

Uatschitchun:
Oh, there's DBHelper.java where DB updates are done on uprades

Uatschitchun:
I've even noticed, that MAX_HR isn't written (nor calculated) for DB.LAP.MAX_HR. Value should possibly written in Step.java onStop(). 
But where to calculate from? Could, afaics, be only calculated from DB.LOCATION.HR in scope.LAP?!

gerhardol:
I did not see any swapped uses, the constant names (actual names of the database columns) could well be xxxx and yyyy.
So how can you get other values?

If possible avoid database upgrades, it makes it much harder to downgrade.
Especially when developing.

Uatschitchun:
This is the DB from my phone:
Screenshot_20211007_155402

MaxHR < AvgHR? ;)

gerhardol:
Adding a comment for this in #1072

gerhardol:
Closing this PR, the TODO is in the code and may be changed at next db update

Issue #1052, opened 2021-10-04
Mapbox & telemetry


Uatschitchun:
Mapbox is sending telemetry data when being used. One is able to "Opt-Out" of this only from the Info-Button on the map. I've read that with Iphones, the mapbox sdk is even sending telemetry data, when the app isn't running, bust just also with "region changes" resulting in volumina of 30-100MB/day.

There's just this:

> Telemetry opt out
> Mapbox Telemetry is a powerful location analytics platform included in this SDK. By default, the SDK sends anonymized location and usage data to Mapbox whenever the host app causes it to be gathered. The Mapbox Terms of Service require your app to provide users with a way to individually opt out of Mapbox Telemetry, which is provided automatically as part of the attribution control. If you hide the attribution control, you must provide an alternative opt out for your users to use.

https://docs.mapbox.com/help/glossary/telemetry/

There's nothing to be found on how to implement this Opt-Out without using the 'attribution'. (:me haven't found anything).

The Info-Button on the map offers to disagree and opt-out of telemetry, but users of RunnerUp do not know about, are not informed and have no prominent place in the app to opt-out. 

RU should inform about Mapbox' telemetry and how to opt-out!

gerhardol:
https://github.com/jonasoreland/runnerup/blob/master/app/listings/play/privacy.html
Is listed in Play page. But Telemetry should be mentioned explicitly.
I did not find any other way to disable telemetry by default other than purchasing a plan (it is using the free version now, with a caps on usage).
This is one reason why MapBox is not in the F-Droid version.

There could be a popup first time opening MapBox (unless telemetry kicks of when starting first time, more complicated to deactivate then.)

The "popup seen" must not be backed up.

Uatschitchun:
If I get it correctly, the option to disable telemetry is in the attribution (copyright info) only. If hiding it, one needs to allow opt-out in another place. Only information/code examples I have found was for iPhones.
Mapbox seems to be quite aggressive in using this data and even acquiring it, when the app itself is not running.
Remember the pop-up you talked about regarding location tracking in one of my PRs?

gerhardol:
> If I get it correctly, the option to disable telemetry is in the attribution (copyright info) only. If hiding it, one needs to allow opt-out in another place. Only information/code examples I have found was for iPhones. Mapbox seems to be quite aggressive in using this data and even acquiring it, when the app itself is not running. 

[...]

PR #1057, opened 2021-10-06
Wip: Update 'About' section

Uatschitchun:
Add Privacy information and info on how to opt-out of Mapbox' telemetry. See: #1052

### DIFF
--- a/app/assets/about.html
+++ b/app/assets/about.html
@@ -4,6 +4,16 @@
 <title>About RunnerUp</title>
 </head>
 <body>
+<h2>Privacy</h2>
+<p>RunnerUp records location (GPS) when an activity is started, as well steps (for cadence), temperature & pressure from internal sensors, if activated.
+    <br>The recording is done also when the app is in background (a notification is always shown when location is enabled)
+    <br>A user can export this information to accounts (files or web services), the information is not shared in any other way.
+</p>
+<p>The external accounts have separate privacy policies.</p>
+<p><b>Note:</b> MapBox enables tracking by default (when using the map to view activities).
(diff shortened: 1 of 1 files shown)


app/assets/about.html
```diff
@@ -4,6 +4,16 @@
...
+<h2>Privacy</h2>
+<p>RunnerUp records location (GPS) when an activity is started, as well steps (for cadence), temperature & pressure from internal sensors, if activated.
+    <br>The recording is done also when the app is in background (a notification is always shown when location is enabled)
+    <br>A user can export this information to accounts (files or web services), the information is not shared in any other way.
+</p>
+<p>The external accounts have separate privacy policies.</p>
+<p><b>Note:</b> MapBox enables tracking by default (when using the map to view activities).
+<br>This can be disabled (opt-out) from the MapBox logo's info button on the lower left corner when viewing the map.
+</p>
+<p>The usage of the map has a separate policy: <a href="https://www.mapbox.com/legal/privacy">MapBox privacy policy</a></p>
```
gerhardol, line 16: The privacy policy in itself should be updated with telemetry etc (maybe cut from here).
The "always using location" notification should be explained too.

Thoughts: All in the app is translated but not this. Probably OK.

gerhardol (commented):
+1

app/assets/about.html
Uatschitchun: For to translate, it doesn't work with strings.xml, I guess!?
Adding language specific html isn't an option, is it?

gerhardol: I believe Transifex could handle this by adding file types, have not investigated.
But that is separate, not part of the PR

Similar with replacing MapBox

PR #1061, opened 2021-10-08
Final: Catch back button to not exit RU accidentally

Uatschitchun:
This is the final version of #1053.

Oftenly, when navigating within RU, one hits back button once too often and exits RU accidentally.

With this PR, exiting RU is only possible from the first Tab (StartActivity) with a double back button press.
All other Tabs (History, Feed, Settings) jump back to first Tab instead.

Pls test and review


gerhardol:
Thanks!
Pushed a change with reformatting the code and fixed the Git commit message
Sorry that it took so long time for this

Issue #1046, opened 2021-09-28
tcx export not a valid garmin tcx v2 ?


zebeuz:
Hello,

I would like to export my activities  to the TCX format and import them in pytrainer database (pytrainer is a free software to store and analyse yours performances). But the tcx from Runner Up is not recognized by pytrainer.

I tried to create a plugin in pytrainer to import TCX files from runner Up but in fact, the problem seems to be a bug in the tcx file with the tag <Creator>.

Runner up mark : 
<Creator xsi:type="Device_t">
<Name>RunnerUp U FEEL LITE</Name>
</Creator>

But the XLM schema descriptor (https://www8.garmin.com/xmlschemas/TrainingCenterDatabasev2.xsd) requires
	<xsd:complexType name="Device_t">
		<xsd:annotation>
			<xsd:documentation>Identifies the originating GPS device that tracked a run or
                               used to identify the type of device capable of handling
                               the data for loading.</xsd:documentation>
		</xsd:annotation>
		<xsd:complexContent>
			<xsd:extension base="AbstractSource_t">
				<xsd:sequence>
					<xsd:element name="UnitId" type="xsd:unsignedInt"/>
					<xsd:element name="ProductID" type="xsd:unsignedShort"/>
					<xsd:element name="Version" type="Version_t"/>
				</xsd:sequence>
			</xsd:extension>
		</xsd:complexContent>
	</xsd:complexType>

A test on online xml validator confirm the problem.

If I add unitId/ProducId/Version like this : 
<Creator xsi:type="Device_t">
        <Name>Runner Up U FEEL LITE</Name>
        <UnitId>3842416961</UnitId>
        <ProductID>1264</ProductID>
        <Version>
          <VersionMajor>2</VersionMajor>
          <VersionMinor>50</VersionMinor>
          <BuildMajor>0</BuildMajor>
          <BuildMinor>0</BuildMinor>
        </Version>
      </Creator>

The file is correct (Online XML schema validator) and pytrainer accept to  import the activity.

PS: Yes, I try to improve my english.

gerhardol:
That was a picky parser...
The added data looks like it is taken from a GF305 with Garmin Training Center.
The values should probably be more generic.

Do you want to send a PR?

zebeuz:
Probably  picky parsers (three). yes. Runner Up start the file with : 
```
[...]

PR #386, opened 2016-05-02
* Possible fix for BUG tcx schema is wrong #332 (part 1).

rickyepoderi:
The other part was already fixed by Jonas Oreland.

Issue #332, opened 2015-09-24
tcx schema is wrong


rickyepoderi:
Hi,

I was playing with your TCX exports and it seems that there are some minor errors according to the garming schema: http://www8.garmin.com/xmlschemas/TrainingCenterDatabasev2.xsd

1.- AverageHeartRateBpm and MaximumHeartRateBpm in the Lap element are situated in a wrong order (they should be after calories and they appear after Tracks in my case).

2.- HeartRateBpm sometimes is 0 in Tracks (because of a read error of something) and the XSD defines it as positiveByte (>0). So you have to put 1 or, even better, omit the HeartRateBpm in this Track beacuse was an error.

I was parsing a file with PHP and the schema validation gave an exception. For the moment I won't check schema. ;-) I think this will be easy to fix.

Thanks!

jonasoreland:
bahh...fixing 1 turned out to not super trivial...
cause average/max is actually computed while creating track :(

fixed 2 though.

i think i'll close this anyway,
reopen if you want to...

rickyepoderi:
Thank you Jonas! But for (1) what about using SQL to get the average and the max? You just do something like "SELECT AVG(hr_col), MAX(hr_col) from table" (maybe where for avoid null) before and then you process the rest as always.

rickyepoderi:
Hi Jonas,

I have implemented the previous idea and it seems to work for me (sorry but not very experienced with android-studio). I have added another method to write the average and maximum heart-rate and I commented out your calculations (better delete the lines if you think my idea is valid). 

```
diff --git a/app/src/org/runnerup/export/format/TCX.java b/app/src/org/runnerup/export/format/TCX.java
index 3711b8f..3156c9f 100644
--- a/app/src/org/runnerup/export/format/TCX.java
+++ b/app/src/org/runnerup/export/format/TCX.java
@@ -142,6 +142,32 @@ public class TCX {
         }
     }
 
+    private void exportLapHeartRate(long activityId, long lapId) throws IOException {
+        Cursor c = mDB.rawQuery("SELECT AVG(" + DB.LOCATION.HR + "), MAX(" + DB.LOCATION.HR + ")" +
+                     " FROM " + DB.LOCATION.TABLE +
+                     " WHERE " + DB.LOCATION.ACTIVITY + " = " + activityId +
+                               " and " + DB.LOCATION.LAP + " = " + lapId +
+                               " and " + DB.LOCATION.HR + " is not null", null);
[...]

PR #1062, opened 2021-10-08
Final: Update for TCX & GPX export:

Uatschitchun:
This is the final version of #1054 

* Update KXmlSerializer with version from https://github.com/stefanhaustein/kxml2
* Update TCX.java for to format exported XML with indents and add Decice&Author (pytrainer seeme to need it)
* Fix for #1046 
* Fix for #332
* Import fixes from https://github.com/jonasoreland/runnerup/pull/386
* Fix some warnings regarding TODO
* Set version & build correctly in exported TCX
* Fix xsi:type
* Construct application name in TCX as: "ApplicationID - Version.Name - Version.Code". Might be handy ...
* Add missing tags and remove obsolete ones
* Add xsi:schemaLocation for completeness. Other TCX have it, too

### DIFF
--- a/app/src/main/org/runnerup/export/format/TCX.java
+++ b/app/src/main/org/runnerup/export/format/TCX.java
@@ -100,7 +100,7 @@ public class TCX {
                     "http://www.w3.org/2001/XMLSchema-instance");
             mXML.attribute("", "xmlns:xsd",
                     "http://www.w3.org/2001/XMLSchema");
-            mXML.attribute("", "xmlns:ae",
+            mXML.attribute("", "xmlns:ns3",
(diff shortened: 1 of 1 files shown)


Uatschitchun:
Will this not be part of 2.4.3.1?

Uatschitchun:
2.4.3.1?

gerhardol:
An update is needed due to that he search is broken, but there is a bad Target 11 that requires son beta time  So maybe updated in the public release. 

I have been on the limit for sometime with real life and some other projects and this is not so easy to verify.

gerhardol:
Finally got to look at the details, looking at the specs and other implementations.
Rebased removing the kml serializer as that was merged prior to this PR.
A few other changes to namespace

PR #1064, opened 2021-10-16
Target Android 11

gerhardol:
A few crash preventions
Android 11 required after Nov 1. Should be working but requires some more steps to enable GPS scanning.


gerhardol:
Done in #1074

PR #1071, opened 2021-11-07
Update strings.xml

sdebaerd:
Added some missing translated texts in Dutch

### DIFF
--- a/common/src/main/res/values-nl/strings.xml
+++ b/common/src/main/res/values-nl/strings.xml
@@ -14,6 +14,7 @@
   ~
   ~  You should have received a copy of the GNU General Public License
   ~  along with this program.  If not, see <http://www.gnu.org/licenses/>.-->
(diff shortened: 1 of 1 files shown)


gerhardol:
Can you please add these on Transifex?
See https://github.com/jonasoreland/runnerup/wiki

There are some changes i master not pushed to sources yet, as that will remove some strings slightly changed English strings where the translations are mostly OK.

sdebaerd:
Overlooked it, my apologies. I signed up and requested to join the projectteam.

sdebaerd:
Added the translations

gerhardol:
I had prepared a 2.4.3 release, will update it with these translations too. 
To be released as a beta tonight.
The paired/non-paired BLE handling maybe need to be separated before a public release, otherwise I expect a week for the public release.

Issue #1084, opened 2021-12-28
TTS Installed but not detected


vgiannoul:
Hello and thanks for your work!
It seems that after v2.4.1.0 TTS is not recognized even though it's installed and working on other apps. The audio cues silently fail when tested via the settings. When I downgrade the app, audio cues are working as expected. 
I'm testing this on a pixel 3a phone running latest grapheneOS  (android 12)

gerhardol:
There are no direct changes to TTS at all, I can only see a few library updates as possibly related (Android 11 targeting in 2.4.2).
I have not seen any changed recommendations for Android 12 or that the implementation is not as recommended, so no clues here.
Any printouts in the log? (amaybe including "res == ERROR emit()").
Any possibility that you can debug?
(add printouts to app/src/main/org/runnerup/view/AudioCueSettingsActivity.java OnInitListener and RUTextToSpeech emitt()).

vgiannoul:
I was able to install the latest app version and get some logs via adb logcat but I'm not sure what is going on.
Still no audio cues and on the same device the older version works as expected.
If you could assist me I could provide more info. 
Thanks again for your time

`01-04 15:09:46.764  1143 10085 W InputManager-JNI: Input channel object '39fc8c8 org.runnerup/org.runnerup.view.AudioCueSettingsActivity (client)' was disposed without first being removed with the input manager!
01-04 15:09:48.984 30349 30349 E org.runnerup.workout.feedback.RUTextToSpeech: setLanguage(English) => MISSING: -2
01-04 15:09:48.986 30349 30349 E org.runnerup.workout.feedback.RUTextToSpeech: buffer speak: interval
01-04 15:09:48.987 30349 30349 E org.runnerup.workout.feedback.RUTextToSpeech: buffer speak: 386 meters
01-04 15:09:48.989 30349 30349 E org.runnerup.workout.feedback.RUTextToSpeech: buffer speak: 4 minutes 49 seconds
`

rickyepoderi:
[...]

PR #1085, opened 2022-01-05
issue #1084: TTS Installed but not detected

rickyepoderi:
Issue: #1084 

Just adding the TTS_SERVICE to the app `AndroidManifest.xml` and two minor modifications to control better the error with the TTS object.

This works for me in lineageos 18.1 / Android 11.

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -43,6 +43,12 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" android:minSdkVersion="28"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
 
+    <queries>
+        <intent>
+            <action android:name="android.intent.action.TTS_SERVICE" />
+        </intent>
+    </queries>
+
     <application
         android:allowBackup="true"
         android:appCategory="maps"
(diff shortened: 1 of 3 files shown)


gerhardol (approved):
Looks good.
Did you check the error handling for normal activities too?
(status is checked there and the tracker state is set to error, so it looks OK to me).

rickyepoderi:
> Did you check the error handling for normal activities too?

I just saw the status was checked in the `TrackerTTS` callback, not looked into that deeply. But I can say that now it works for me in running activities too, it speaks at start, stop, resume and the lap summary (only tested in my real device, that's what I have configured there). In general it was running OK before in Android 10 so I think just the xml stanza is really needed, although I also added the two others checks for completeness.

gerhardol:
> > Did you check the error handling for normal activities too?
> 
> I just saw the status was checked in the `TrackerTTS` callback, not looked into that deeply. But I can say that now it works for me in running activities too, it speaks at start, stop, resume and the lap summary (only tested in my real device, that's what I have configured there). In general it was running OK before in Android 10 so I think just the xml stanza is really needed, although I also added the two others checks for completeness.

So it just did not work before, there should be a printout like `state.set(ERROR`
Leave it at this, will prepare a release. (even if Pebble is not resolved it seems)

rickyepoderi:
Ah! Yes, it didn't speak at all since I upgraded the device to lineageos 18.1 / android 11. The activity was recorded OK but without TTS. :smile:

Thanks for looking this!

PR #1090, opened 2022-02-09
Update Formatter.java to fix month display on some languages

aseques:
Hi, there's an issue in the ICU notation format that makes some languages to look bad in the history screen of runner up, for example in catalan the history appears as "de gen. 2022" (there's an extra article) instead of "gen. 2022" that would be the proper text in this case.
Using LLL instead of MMM fixes this for the affected languages (Belarusian, Catalan, Croatian, Finnish, Greek, Lithuanian, Polish, Russian,
Slovak, Ukrainian and several more...) without touching the rest. (see the explanation in here https://developer.android.com/reference/android/icu/text/SimpleDateFormat)

image

### DIFF
--- a/app/src/main/org/runnerup/util/Formatter.java
+++ b/app/src/main/org/runnerup/util/Formatter.java
@@ -79,7 +79,7 @@ public class Formatter implements OnSharedPreferenceChangeListener {
 
         dateFormat = android.text.format.DateFormat.getDateFormat(ctx);
         timeFormat = android.text.format.DateFormat.getTimeFormat(ctx);
-        monthFormat = new SimpleDateFormat("MMM yyyy", cueResources.defaultLocale);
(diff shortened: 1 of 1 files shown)


gerhardol:
Seem reasonable

knarrff:
Since in the text above, the difference is not really visible, maybe here a reference to a page where it becomes a little clearer: https://docs.huihoo.com/android/5.0/reference/java/text/SimpleDateFormat.html

> The two pattern characters L and c are ICU-compatible extensions, not available in the RI or in Android before Android 2.3 (Gingerbread, API level 9). These extensions are necessary for correct localization in languages such as Russian that make a grammatical distinction between, say, the word "June" in the sentence "June" and in the sentence "June 10th"; the former is the stand-alone form, the latter the regular form (because the usual case is to format a complete date). The relationship between E and c is equivalent, but for weekday names.

gerhardol (approved):

PR #1102, opened 2022-04-17
Fix manual entry of time

### LINKED ISSUE #980: The start hour entered in the manual entry menu is increased by an hour when saved
Here i entered 15:30:
Screenshot_20201022-160711
In the history it shows 16:30:
Screenshot_20201022-160718

knarrff:
Do not use (new Date()) to get the timezone offset from, but the date
that was actually entered: d. This fixes the problem that the entered
time was later reported as one hour later when DST was active at the
current location.

This should fix #980.


gerhardol:
Could work

knarrff:
I have changed the request even more after I realized that the old approach still had problems across different time zones, e.g. if "today" is within DST, but the entered date was not. The new approch deals with that and should be even more readable than the old.

gerhardol (approved):

gerhardol:
thanks

PR #1103, opened 2022-04-17
Prevent errors in case numeric settings are empty

### LINKED ISSUE #1100: no recording when gps recording distance is empty
Hi,

I used the app just fine for a few times and it worked ok. Then I started playing around with GPS recording settings to get the recording a little smoother. One of the things I tried was setting a "GPS distance"; it was empty to begin with. This did not improve accuracy, so I removed the value again (opened the dialog, removed the number, tapped on "ok": the field was now empty again).

However: with this now "empty" setting, while I can start a recording, it does not actually record anything else besides the start time (and it only shows that in h:m and not even seconds). No track and no actual time is recorded: I end up with 0km and 0s: no data.

When I go again into settings and enter something for the GPS distance recording setting, like 10m, recording works fine again. This is a little surprising, as the setting was shown as "empty" initially as well and worked fine. Also, setting this to "empty" should not disable recording, but only use the "time" version of that setting.

Anyway: thanks for all the work on the app!

knarrff:
Even though some of the settings like pollDistance and pollInterval are
set to be numeric and (at least as I tested) cannot be negative or
contain anything else than numbers, they can be an empty string. Parsing
that empty string to parseInt() is causing trouble.

This patch looks for empty strings and in that case uses the default
value instead. In addition, it resets the preference value to that
default to make the usage of that default clear to the user.

This should close #1100.

Note that this only happens here when those values are used: when an
activity starts. Ideally, this should be prevented by Android itself,
but barring that, it should be checked right after the setting is
changed in the settings dialog. However, this is beyond my current
Android-Voodoo.

Also note, that this code (like the original version) duplicates the
default values from the file settings.xml. Ideally, that default should
be obtained and used instead - alas I do not know how that can be done.

### DIFF
--- a/app/src/main/org/runnerup/tracker/component/TrackerGPS.java
+++ b/app/src/main/org/runnerup/tracker/component/TrackerGPS.java
@@ -24,6 +24,7 @@ import android.location.Location;
 import android.location.LocationManager;
 import android.os.Handler;
 import android.preference.PreferenceManager;
+import android.text.TextUtils;
 
 import androidx.core.content.ContextCompat;
 
@@ -73,6 +74,17 @@ public class TrackerGPS extends DefaultTrackerComponent implements TickListener 
         return ResultCode.RESULT_OK;
     }
 
+    private Integer parseAndFixInteger(int resId, String def, Context context) {
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
+        String s = preferences.getString(context.getString(resId), def);
+        if (TextUtils.isEmpty(s)) {
+            SharedPreferences.Editor prefedit = preferences.edit();
+            prefedit.putString(context.getString(resId), def);
+            prefedit.apply();
+            return Integer.parseInt(def);
+        }
+        return Integer.parseInt(s);
+    }
     @Override
     public ResultCode onConnecting(final Callback callback, Context context) {
         if (ContextCompat.checkSelfPermission(this.tracker,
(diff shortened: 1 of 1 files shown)


gerhardol:
Seem OK, will go into details later

> default values from the file settings.xml. Ideally, that default should
be obtained and used instead - alas I do not know how that can be done.

Agree

app/src/main/org/runnerup/tracker/component/TrackerGPS.java
```diff
@@ -73,6 +74,17 @@ public ResultCode onInit(final Callback callback, Context context) {
...
     }
 
+    private Integer parseAndFixInteger(int resId, String def, Context context) {
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
+        String s = preferences.getString(context.getString(resId), def);
+        if (TextUtils.isEmpty(s)) {
+            SharedPreferences.Editor prefedit = preferences.edit();
+            prefedit.putString(context.getString(resId), def);
+            prefedit.apply();
+            return Integer.parseInt(def);
```
gerhardol, line 84: In this case the early return is just confusing
```suggestion
            s = def;
```

app/src/main/org/runnerup/tracker/component/TrackerGPS.java
```diff
@@ -83,14 +95,12 @@ public ResultCode onConnecting(final Callback callback, Context context) {
         try {
             LocationManager lm = (LocationManager) context.getSystemService(Context.LOCATION_SERVICE);
             SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
```
gerhardol, line 97: Either remove or pass to parseAndFixInteger

gerhardol (approved):

PR #1104, opened 2022-04-18
Added support for Osmdroid in the F-Droid version of RunnerUp

Upabjojr:
The `MapView` widget will now be visible in the free version of RunnerUp. It will link to an Osmdroid widget which is compatible with F-Droid.

This edit allows to view the track on a map with the free version on F-Droid.

The play version will remain identical to the current one.

### DIFF
--- a/app/src/free/org/runnerup/util/MapWrapper.java
+++ b/app/src/free/org/runnerup/util/MapWrapper.java
@@ -24,7 +24,6 @@ import android.os.AsyncTask;
 import android.os.Bundle;
 
 import org.osmdroid.api.IMapController;
-import org.osmdroid.api.IMapView;
 import org.osmdroid.config.Configuration;
 import org.osmdroid.tileprovider.tilesource.TileSourceFactory;
 import org.osmdroid.util.GeoPoint;
@@ -32,6 +31,7 @@ import org.osmdroid.views.CustomZoomButtonsController;
 import org.osmdroid.views.MapView;
 import org.osmdroid.views.overlay.Marker;
 import org.osmdroid.views.overlay.Polyline;
+import org.runnerup.R;
 import org.runnerup.common.util.Constants;
 import org.runnerup.db.entities.LocationEntity;
 
@@ -39,6 +39,8 @@ import java.util.LinkedList;
 import java.util.List;
 
 public class MapWrapper implements Constants {
+
+    public static final boolean USING_OSMDROID = true;
 
     private final Context context;
     private final SQLiteDatabase mDB;
@@ -97,6 +99,7 @@ public class MapWrapper implements Constants {
             long mID = params[0].mID;
             IMapController iMapController = params[0].iMapController;
             Polyline route = new Polyline(mapView, true);
+            route.setInfoWindow(null);
             route.getOutlinePaint().setStrokeWidth(10.f);
             LocationEntity.LocationList<LocationEntity> ll = new LocationEntity.LocationList<>(mDB, mID);
             List<GeoPoint> points = new LinkedList<>();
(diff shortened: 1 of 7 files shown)


app/src/main/org/runnerup/view/DetailActivity.java
```diff
@@ -127,13 +127,10 @@ public class DetailActivity extends AppCompatActivity implements Constants {
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        if (BuildConfig.MAPBOX_ENABLED > 0) {
```
gerhardol, line 130: This is still needed for Play version without MapBox key.

gerhardol (commented):
Will likely take me some days to get into this...

app/src/main/org/runnerup/view/DetailActivity.java
Upabjojr: What about this:
```java
if (BuildConfig.APPLICATION_ID.contains(".free") || BuildConfig.MAPBOX_ENABLED > 0) {
    ...
}
```
Is there a better way to understand if the code was compiled using the free version?

gerhardol:
Seem to be working fine.
I got some exception at first use but before I could investigate the logs were gone.
The markers could be improved.
Also, if no points at all the zoom level is max, compared to min with MapBox.
This could well be replacing the MapBox inplementation also in Play. In addition to the free aspect, the map usage is limited in MapBox.

Upabjojr:
> The markers could be improved.

Osmdroid appears to have a very limited choice of icons for markers. External icons may be needed.

> This could well be replacing the MapBox inplementation also in Play. In addition to the free aspect, the map usage is limited in MapBox.

Bear in mind that this would use OpenStreetMap's servers in the Play version. The servers are run by donations, while MapBox is a for-profit company. Not sure we should abuse OpenStreetMap's servers while there are alternatives.

knarrff:
> Not sure we should abuse OpenStreetMap's servers while there are alternatives.

I doubt usage of OSM in RunnerUp would even show up in their overall usage. Also, why would it be abuse to use OSM here - wasn't is built to be used?

Upabjojr:
So do you suggest to switch to Osmdroid by default and drop MapBox?

knarrff:
This is not up to me: I am not even a developer. I just found the reasoning a little strange in "Not sure we should abuse OpenStreetMap's servers while there are alternatives."

gerhardol:
The primarily reason for RunnerUp is that is an open source application. That is the reason why I maintain it - I am not really a user myself but I want to keep alternatives open.
(I recently broke my new phone when bring it for a run).
It is occasionally fun to develop, even if I have got turned off by all changes Google enforces.

So just using free SW is in line with my intention. I am am not strict about it, there are hidden upload keys too. The MapBox usage is not very high. 5K users in Play, the free MapBox account has currently 634/25000 monthly users.
Usage in F-Droid is unknown, I have no statistics at all there.

For now we can keep MapBox in Play, may drop it at next API change they make.

Upabjojr:
> Usage in F-Droid is unknown, I have no statistics at all there.

I guess there are probably less users on F-Droid than on Play. I don't have Play on my phone so I use the F-Droid version, but I image most people do install the app from Play.

app/src/free/org/runnerup/util/MapWrapper.java
```diff
@@ -17,35 +17,135 @@
...
 import org.runnerup.common.util.Constants;
-import org.runnerup.util.Formatter;
+import org.runnerup.db.entities.LocationEntity;
 
+import java.util.LinkedList;
+import java.util.List;
 
 public class MapWrapper implements Constants {
 
+    public static final boolean USING_OSMDROID = true;
```
gerhardol, line 43: Use buildConfigField instead to set this.

app/src/free/org/runnerup/util/MapWrapper.java
```diff
@@ -17,35 +17,135 @@
...
+                markerEnd.setTextIcon(context.getString(R.string.Finished));
+
+                markerStart.setAnchor(Marker.ANCHOR_CENTER, Marker.ANCHOR_BOTTOM);
+                markerEnd.setAnchor(Marker.ANCHOR_CENTER, Marker.ANCHOR_BOTTOM);
+
+                mapView.getOverlays().add(markerStart);
+                mapView.getOverlays().add(markerEnd);
+            }
+
+            mapView.getOverlays().add(route);
```
gerhardol, line 132: Should this be added only if there are points?

gerhardol (commented):
Just a few comments

Upabjojr:
I think this feature should be added to the F-Droid version at first, as I'm not sure it's as complete as the MapBox widget. This can still be changed in the future.

gerhardol (approved):

Issue #1105, opened 2022-04-22
.tcx file is empty (0KB) for manual activities (no gps data)


Hirutenshi:
Sharing or downloading (to file) a manually added activity (without any gps tracking data) creates an empty (0KB) tcx file.

knarrff:
I can confirm.

gerhardol:
RU is is only designed to be used with GPS. Similar, recording is noly done when GPS is updated.
Manual registration of activities is only implemented really for the local history.

knarrff:
I can't see why exporting the metadata should not work in principle. It does for GPX, and that code looks similar in nature. I don't have time right now to look into the issue. Given the zero bytes despite the assignment of the general metadata initially in the exporter I suspect an exception somewhere in that code, leaving the "share" with an empty file instead of one with at least the manually created metadata in it.

> RU is is only designed to be used with GPS.

I agree that using RU essentially as a glorified stop watch isn't the main intention. However, why not? There might be users with reasons they find valid. One of them could be that sometimes GPS recording just does not work or not work well. Or you simply forgot your phone but want to have at least the distance/time data all in one place: RU on your phone, so you add the data manually. Later you might want to export all data. So, with manual entries existing, why shouldn't you be able to export them as well.

gerhardol:
As the intention (before I was involved) was just GPS, this was just not implemented, can for sure be done.
No GPS is just not implemented.

Hirutenshi:
> I agree that using RU essentially as a glorified stop watch isn't the main intention. However, why not? There might be users with reasons they find valid. One of them could be that sometimes GPS recording just does not work or not work well. Or you simply forgot your phone but want to have at least the distance/time data all in one place: RU on your phone, so you add the data manually. Later you might want to export all data. So, with manual entries existing, why shouldn't you be able to export them as well.

For the past 2 years I have tracked all my running, biking, hiking and rowing activities using RU. Having all my activities in one app is a wonderful thing. Since I use a rowing machine there is no gps tracking for these activities. It never occured to me that this might be a problem if I ever intend to export my activities. 

If you decide that this is not a desired feature a short warning to inform users about this behaviour would be helpful.

gerhardol:
It is a desired feature, just not something intended initially.
If the app is used this way now, it should be added. There is no intention to lock-in to the app.
Similar for custom activities (could be user defined to avoid translation issues).
Patches are welcome.

Hirutenshi:
[...]

PR #1107, opened 2022-05-05
Fix misplaced endTag for TCX exports without trackpoints

Hirutenshi:
misplaced endTag causes exceptions for all exports without tracking data.

related issue: #1105

### DIFF
--- a/app/src/main/org/runnerup/export/format/TCX.java
+++ b/app/src/main/org/runnerup/export/format/TCX.java
@@ -347,8 +347,8 @@ public class TCX {
                     mXML.text(formatTime(startTime));
                     mXML.endTag("", "Time");
                     mXML.endTag("", "Trackpoint");
+                    mXML.endTag("", "Track");
                 }
-                mXML.endTag("", "Track");
 
                 if (cntHR > 0) {
                     mXML.startTag("", "AverageHeartRateBpm");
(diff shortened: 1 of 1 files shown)


app/src/main/org/runnerup/export/format/TCX.java
```diff
@@ -347,8 +347,8 @@ private void exportLaps(long activityId, long startTime, Sport sport) throws IOE
                     mXML.text(formatTime(startTime));
                     mXML.endTag("", "Time");
                     mXML.endTag("", "Trackpoint");
+                    mXML.endTag("", "Track");
                 }
-                mXML.endTag("", "Track");
```
gerhardol, line 351: This is where end tag is added for the normal case, so it will break the standard export
addGratuitousTrack is always false

Something like the following could work
```
                if (!hasTrackpoints && addGratuitousTrack) {
                    mXML.startTag("", "Track");
                    mXML.startTag("", "Trackpoint");
                    mXML.startTag("", "Time");
                    mXML.text(formatTime(startTime));
                    mXML.endTag("", "Time");
                    mXML.endTag("", "Trackpoint");
                    mXML.endTag("", "Track");
                }
                else if (hasTrackpoints) {
                    mXML.endTag("", "Track");
                }
```

Hirutenshi: oh, you are right. Was so focused on the edge case that I missed the default case... I tested your suggestion and was able to export activities with gps tracking as well as without. Thank you.

gerhardol (approved):

PR #1114, opened 2022-06-09
Remove feed implementation

gerhardol:
Since Endomondo was removed, no service supports Feed implementation (activities for friends).
This removes the tab and implementation
DB is not updated, to be done when there is another change in the db.

Workouts still had a non functional Download button


gerhardol:
Moving the "date handling" to another PR

PR #1118, opened 2022-08-19
Fix Farsi language label

gigilibala:
There is no need for the word "zaban" which means language itself. Just
"Farsi" is the correct use.

### DIFF
--- a/app/res/values/array.xml
+++ b/app/res/values/array.xml
@@ -45,7 +45,7 @@
     <item>Čeština</item>
     <item>Deutsch</item>
     <item>Español</item>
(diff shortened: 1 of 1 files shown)


gerhardol (approved):
I believe I copied this from wikipedia and asked someone that speaks Farsi but this obviously was not correct despite this...

gigilibala:
> I believe I copied this from wikipedia and asked someone that speaks Farsi but this obviously was not correct despite this...

Thanks for approving this. Yeah, there is an extra word in there that means "language" and is not necessary. You can probably double check this under language settings of Android too.

PR #1131, opened 2023-02-09
App strings reworked

comradekingu:


gerhardol:
I pushed some changes to align words like "heart rate" and "auto lap" to what I found as most common at quick searches and to what Garmin uses.
Other aligned to the names used in Android. Except for text-to-speech, hyphens were removed (including a few existing before).

A few other changes cannot be changed like on line 289/290, Google would not allow publish in Play without that wording (the message is not right, I wanted to have something including background and notification, but had to give up).

comradekingu:
@gerhardol What is the point of that commit? All of that is worse.
It doesn't observe any of the rules for dashes or brackets.
Moreover, some of the most egregious sentences are back.

gerhardol:
> @gerhardol What is the point of that commit? All of that is worse. It doesn't observe any of the rules for dashes or brackets. Moreover, some of the most egregious sentences are back.

See explanation above. 

"heart-rate" seem less common than "heart rate" for instance.

Also, if merging, the existing translations will be pushed, this affects English, to very little extent the translations.

comradekingu:
It isn't just "heart-rate" though, it is heart-rate ___(something)__, or ___(something)___ heart-rate.

I managed to not get the list of similar projects on Hosted Weblate posted, but my idea was to move the translation effort there. That way translations are only fuzzified, and not lost when anything changes.

gerhardol:
> It isn't just "heart-rate" though, it is heart-rate _**(something)**, or _**(something)**_ heart-rate. (I added some explanation to my above reply.)

But that is a common problem in English. I do not find that as an argument to add hyphen where it is normally not used.
I preferred heartrate as one word, that is in line with my native language. Similar for autolap and autopause.

Some of the examples you have above should be changed though. I reverted those that changed the wording in Android, like "Nearby devices permission", no hyphen used.
"Granting location permission required" should have 'permission' in the string, it is what the user see.
But I am both overworked and recovering after an accident and have other projects, I have limited time to update proposals.

> I managed to not get the list of similar projects on Hosted Weblate posted right away, but my idea was to move the translation effort there. That way translations are only fuzzified, and not lost when anything changes. I don't want my changes in Transifex.

I still fear that more translators are lost than gained by features in other platforms.

Translations in the app should probably be limit to those with good coverage, all are used now.
This follows the the policy of the original author (and designer of most functionality including strings).
I am just a maintainer that basically just use the app to test changes. Most time is spent on changes required by Google. But I want a free alternative to all closed source apps. Lack of new features is what is the highest risk for the app, it is loosing users.

There has been no feedback at all for 2.5.0. Will wait a week or so more before updating in Play.

common/src/main/res/values/strings.xml
```diff
@@ -237,7 +237,7 @@
   <string name="Current">Current</string>
   <string name="Misc_cues">Other cues</string>
   <string name="Target_coaching">Target coaching</string>
-  <string name="Skip_event_audio_cues">Skip event audio cues</string>
+  <string name="Skip_event_audio_cues">Skip audio cues for event</string>
```
jdicke, line 240: I liked Skip event audio cues better

comradekingu: That is just word salad. Should be "events", but alas.
I don't want to contribute to Transifex, so lets just sink this effort.

PR #1139, opened 2023-03-20
Use timers that take account of device sleep

PhilArmstrong:
System.nanoTime() does not measure time correctly across device sleep periods.

Use SystemClock.elapsedRealtime() & SystemClock.elapsedRealtimeNanos() instead.

### DIFF
--- a/app/src/main/org/runnerup/tracker/Tracker.java
+++ b/app/src/main/org/runnerup/tracker/Tracker.java
@@ -99,7 +99,7 @@ public class Tracker extends android.app.Service implements
     private final TrackerElevation trackerElevation = (TrackerElevation) components.addComponent(new TrackerElevation(this, trackerGPS, trackerPressure));
     TrackerReceiver trackerReceiver = (TrackerReceiver) components.addComponent(new TrackerReceiver(this));
     private TrackerWear trackerWear; // created if version is sufficient
-    private TrackerPebble trackerPebble; // created if version is sufficient
+    private TrackerPebble trackerPebble; // created if version is sufficient>
 
     private boolean mBug23937Checked = false;
     private long mSystemToGpsDiffTimeNanos = 0;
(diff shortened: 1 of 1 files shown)


PhilArmstrong:
Sorry about the stray comment edit there.

app/src/main/org/runnerup/tracker/Tracker.java
```diff
@@ -644,20 +644,20 @@ private void onLocationChangedImpl(Location arg0, boolean internal) {
         if (!mTimeFromGpsPoints || internal) {
             // Set internal time also if mBug23937Checked is not set
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
-                long now = System.nanoTime();
+                long now = SystemClock.elapsedRealtimeNanos();
                 arg0.setElapsedRealtimeNanos(now - mSystemToGpsDiffTimeNanos);
             } else {
-                long now = System.currentTimeMillis();
```
gerhardol, line 650: Cleanup is needed here, mixing elapsed and absolute time.
So some more thoughts is needed

app/src/main/org/runnerup/tracker/Tracker.java
```diff
@@ -644,20 +644,20 @@ private void onLocationChangedImpl(Location arg0, boolean internal) {
         if (!mTimeFromGpsPoints || internal) {
             // Set internal time also if mBug23937Checked is not set
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
-                long now = System.nanoTime();
```
gerhardol, line 647: The only chanhe in this PR is line 647 and 656,
System.currentTimeMillis() is handled OK.

However, the handling of old versions should be cleaned up,  maybe setting minimum version to 6 or so, about 46 would be affected - just over 1% of users.

I pushed xhanges to a private branch, will you adopt those or should I make a new PR?

gerhardol:
Included with the proposed changes in #1141 
Thanks!

Issue #648, opened 2018-03-23
Replace TabActivity with AppCompatActivity


12people:
TabActivity has long been deprecated, and its use prevents niceties from the support library—vector drawables, for example.

Please use AppCompatActivity for MainLayout and fragments for its individual components (StartActivity, FeedActivity, HistoryActivity, SettingsActivity).

All of this is supported on Froyo, so there should be no compatibility issues.

gerhardol:
A partial conversion to AppCompat is available in branch: https://github.com/gerhardol/runnerup/tree/feature/appcompat-changes

gerhardol:
Migration to androidx and AppCompatActivity done where possible. Only TabActivity and SettingsActivity remains, which is quite some changes.

roberi:
> Only TabActivity and SettingsActivity remains, which is quite some changes.

I have been working on migrating all settings to AndroidX Preferences. Before I make a PR I wonder if there is some special reason why the SettingsActivity has no Toolbar (app bar)? Both StartActivity and HistoryActivity have one.

![image](https://github.com/jonasoreland/runnerup/assets/10847866/d9f38518-3a73-42f7-89fe-3bdc2e3da817)

Should we add a Toolbar to SettingsActivity as well?

gerhardol:
> > Only TabActivity and SettingsActivity remains, which is quite some changes.
> 
> I have been working on migrating all settings to AndroidX Preferences. Before I make a PR I wonder if there is some special reason why the SettingsActivity has no Toolbar (app bar)? Both StartActivity and HistoryActivity have one.

> Should we add a Toolbar to SettingsActivity as well?

No specific reason, it just happen to be this way....

The recommended handling for settings is not to have it in the toolbar. But with only two other tabs, it is not a space issue.

Looking forward for this - I have gotten stuck every time I started.

roberi:
> The recommended handling for settings is not to have it in the toolbar.

I'm not sure I understand. Or perhaps we misunderstand each other? Why not like in the right picture?

![image](https://github.com/jonasoreland/runnerup/assets/10847866/1d7e4fed-e80a-47ac-b500-da8353d5d049)

> Looking forward for this - I have gotten stuck every time I started.

Looking forward for your feedback. =)

gerhardol:
> > The recommended handling for settings is not to have it in the toolbar.
> 
> I'm not sure I understand. Or perhaps we misunderstand each other? Why not like in the right picture?

[...]

PR #1176, opened 2024-03-17
Compile SDK 34

gerhardol:
Prepare for Android 14
onBackPressed() is deprecated


gerhardol:
I have briefly run this on A12 and A9 too. Can maybe make a public beta next weekend.

roberi:
Seems to work on my Pixel 6 (A14) as well.

But one thing to keep in mind in future updates (for example after replacing TabActivity in #648)... For the new "[predictive back animations](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture)" to work as intended, we should disable a OnBackPressedCallback instead of eg. calling finish() in the callback. So eg. in the DetailActivity the callback should only be enabled (intercept back navigation) while uploading.

gerhardol:
> Seems to work on my Pixel 6 (A14) as well.
> 
> But one thing to keep in mind in future updates (for example after replacing TabActivity in #648)... For the new "[predictive back animations](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture)" to work as intended, we should disable a OnBackPressedCallback instead of eg. calling finish() in the callback. So eg. in the DetailActivity the callback should only be enabled (intercept back navigation) while uploading.

The back navigation is adjusted to go to StartActivity when back in Settings as some other tweaks like disabling back in Run and just disabling GPS when it has been restarted. For instance the last requires some other way to disable GPS.
So this is the fix for updating, not the functionality adopted to A14.
I do not require that the app works as it does now, most changes are improvements...

roberi:
Thanks for the clarification.

gerhardol:
When using this I saw an activity being paused just after starting. That was probably accidentally done when putting away the phone but I want some more usage...

gerhardol:
A few other changes added to this PR

To be able to push an update in Play, Wear must be disabled.

Google Play requires update from the ancient play wear version,
which requires migration to the ancient (non-bundled) wear interface.
The wear app was not built as it was and had to be sideloaded on the watch.
It is possible to revert the patch and download in the phone to continue to use Wear.
(I do not have any Wear device and cannot do the migration.)

PR #1177, opened 2024-04-02
Fix some output streams writing too many bytes

roberi:
There is a bug in the `FileUtil.copy` method where the output stream writes too much data, causing the copied file not to be the same as the original file. This fix ensures that the output stream writes only the number of bytes read from the input stream.

### DIFF
--- a/app/src/main/org/runnerup/util/FileUtil.java
+++ b/app/src/main/org/runnerup/util/FileUtil.java
@@ -29,9 +29,10 @@ public class FileUtil {
     private static int copy(InputStream src, OutputStream dst) throws IOException {
         int cnt = 0;
         byte[] buf = new byte[1024];
-        while (src.read(buf) > 0) {
-            cnt += buf.length;
-            dst.write(buf);
+        int bytesRead;
+        while ((bytesRead = src.read(buf)) > 0) {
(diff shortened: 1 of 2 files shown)


gerhardol (approved):
Potentially the same in Encryption encrypt (only 8 bytes, probably OK) and ManageWorkoutsActivity (saveImport not common)

roberi:
> Potentially the same in Encryption encrypt (only 8 bytes, probably OK) and ManageWorkoutsActivity (saveImport not common)

Yes, definitely the same in ManageWorkoutsActivity. Should I edit this PR and make it more general to include a fix for ManageWorkoutsActivity as well, or make a new PR?

gerhardol:
> > Potentially the same in Encryption encrypt (only 8 bytes, probably OK) and ManageWorkoutsActivity (saveImport not common)
> 
> Yes, definitely the same in ManageWorkoutsActivity. Should I edit this PR and make it more general to include a fix for ManageWorkoutsActivity as well, or make a new PR?

You can add to this PR.
Those extra occurrences are not occurring often, but it is nice to handle them at the same time. (saveImport at least)
I have not checked where this PR change is actually used.

gerhardol (approved):

PR #1190, opened 2024-07-05
Use explicit PendingIntents for notification actions

### LINKED ISSUE #1189: Notification action buttons not working on Android 14
Action buttons in the notifications, like the ones in the picture below, seems not to be working on Android 14.

image

At least not for me on my Pixel 6 running Android 14. They work when I test in Android Studio on an AVD running Android 13.

App version 2.7.0.1

roberi:
Fixes #1189

Starting with Android 14, implicit intents are restricted for internal app components. This commit explicitly sets the package name for PendingIntents used as notification actions to ensure compatibility with the behavior changes outlined in: https://developer.android.com/about/versions/14/behavior-changes-14#security

### DIFF
--- a/app/src/main/org/runnerup/notification/GpsBoundState.java
+++ b/app/src/main/org/runnerup/notification/GpsBoundState.java
@@ -28,6 +28,7 @@ public class GpsBoundState implements NotificationState {
         PendingIntent pi = PendingIntent.getActivity(context, 0, i, intentFlags);
 
         Intent startIntent = new Intent()
+                .setPackage(context.getPackageName())
                 .setAction(Constants.Intents.START_ACTIVITY);
         PendingIntent pendingStart = PendingIntent.getBroadcast(
                 context, 0, startIntent, PendingIntent.FLAG_UPDATE_CURRENT | intentFlags);
--- a/app/src/main/org/runnerup/notification/OngoingState.java
+++ b/app/src/main/org/runnerup/notification/OngoingState.java
@@ -39,11 +39,13 @@ public class OngoingState implements NotificationState {
         PendingIntent pi = PendingIntent.getActivity(context, 0, i, intentFlags);
 
         Intent lapIntent = new Intent()
+                .setPackage(context.getPackageName())
                 .setAction(Constants.Intents.NEW_LAP);
         PendingIntent pendingLap = PendingIntent.getBroadcast(
                 context, 0, lapIntent, PendingIntent.FLAG_UPDATE_CURRENT | intentFlags);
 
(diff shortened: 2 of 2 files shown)


gerhardol:
Approved, working fine

There are two occurrences each in app/src/main/org/runnerup/tracker/component/TrackerPebble.java:85 and app/src/main/org/runnerup/tracker/component/TrackerPebble.java:479
Even if Wear is dropped right now and there are likely a minimum of Pebble users, I suggest that is changed.
I would have expected that app/src/main/org/runnerup/tracker/component/TrackerPebble.java:735 :759 would be affected too

I thought I had checked this already...

roberi:
> There are two occurrences each in app/src/main/org/runnerup/tracker/component/TrackerPebble.java:85...

I agree, this we should change.

But I'm unsure why LocalBroadcastManager is used to broadcast RESUME_WORKOUT and PAUSE_WORKOUT and context.sendBroadcast() is used to broadcast START_WORKOUT.

I think we could either just change startBroadcastIntent.setPackage(null) to startBroadcastIntent.setPackage(context.getPackageName()) or to replace

```java
Intent startBroadcastIntent = new Intent()
    .setAction(org.runnerup.common.util.Constants.Intents.START_WORKOUT);
startBroadcastIntent.setPackage(context.getPackageName());
context.sendBroadcast(startBroadcastIntent);
```
with sendLocalBroadcast(org.runnerup.common.util.Constants.Intents.START_WORKOUT)

Any thoughts about this?

roberi:
> ... and app/src/main/org/runnerup/tracker/component/TrackerPebble.java:479

I assume you mean app/src/wear/org/runnerup/tracker/component/TrackerWear.java:479?

I agree, but same question here. Do we use contextsendBroadcast() or LocalBroadcastManager?

roberi:
> I would have expected that app/src/main/org/runnerup/tracker/component/TrackerPebble.java:735 :759 would be affected too

You must mean some other class than TrackerPebble (since it only has 179 lines)?

gerhardol:
Failed copy...
app/src/main/org/runnerup/view/StartActivity.java:735 :759
app/src/wear/org/runnerup/tracker/component/TrackerWear.java:479

> I think we could either just change startBroadcastIntent.setPackage(null) to startBroadcastIntent.setPackage(context.getPackageName()) or to replace
with sendLocalBroadcast(org.runnerup.common.util.Constants.Intents.START_WORKOUT)

> Any thoughts about this?

No opinion right now. Should try to connect to Pebble users.
Maybe just open an issue about it for now and merge this as is.

PR #1192, opened 2024-08-05
Audio countdown feedback

Emilv2:
Hi,

I'm loving this app, but I had 2 small annoyances:

* When counting down time 10, 5, 3, 2 ,1 just feels so off to me, why skip just 1 number?   10, 5, 4, 3, 2, 1 feels much better to me (I see it's been in the code since 2013, so maybe it's just me).
* I would like to have more countdown feedback when the countdown is longer than 1 min. I doubled the the feedback until it gets bigger than half the original countdown time.

Since they are very small changes I just implemented them directly in a PR

### DIFF
--- a/app/src/main/org/runnerup/workout/WorkoutBuilder.java
+++ b/app/src/main/org/runnerup/workout/WorkoutBuilder.java
@@ -536,6 +536,11 @@ public class WorkoutBuilder {
                 return;
         }
 
+        // Extent the feedback list with more values for longer countdowns
+        while (step.getDurationValue() / 2 > list.get(0)) {
+            list.add(0, list.get(0) * 2d);
+        }
+
         // Remove all values in list close to the step
         while (list.size() > 0 && step.getDurationValue() < list.get(0) * 1.1d) {
             list.remove(0);
--- a/app/src/main/org/runnerup/workout/WorkoutBuilder.java
+++ b/app/src/main/org/runnerup/workout/WorkoutBuilder.java
@@ -521,7 +521,7 @@ public class WorkoutBuilder {
             case TIME:
                 // seconds
                 Double[] tmp0 = {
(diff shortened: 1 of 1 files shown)


app/src/main/org/runnerup/workout/WorkoutBuilder.java
```diff
@@ -521,7 +521,7 @@ private static void createAudioCountdown(Step step) {
             case TIME:
                 // seconds
                 Double[] tmp0 = {
-                        60d, 30d, 10d, 5d, 3d, 2d, 1d
+                        60d, 30d, 10d, 5d, 4d, 3d, 2d, 1d
```
gerhardol, line 524: This is intentional for me, I do not want to be disturbed too much.
This is basically a 1,2,5,10,20,50 sequence modified for 60s and added 3
The 3 could be dropped. 

Someone else have an opinion?
I could live with this.

app/src/main/org/runnerup/workout/WorkoutBuilder.java
```diff
@@ -536,6 +536,11 @@ private static void createAudioCountdown(Step step) {
                 return;
         }
 
+        // Extent the feedback list with more values for longer countdowns
+        while (step.getDurationValue() / 2 > list.get(0)) {
+            list.add(0, list.get(0) * 2d);
```
gerhardol, line 541: or add every minute?
OK for me

Emilv2: I took doubling over every minute because it fits in the sequence and might be less disturbing, but I am fine with either

app/src/main/org/runnerup/workout/WorkoutBuilder.java
Emilv2: I would say dropping the 5 feels more natural than dropping the 3

gerhardol:
thanks

PR #1211, opened 2025-05-15
resurrect wear

jonasoreland:
1) Remove usage of NodeApi,
it mostly works without it. And for the scenario
where the phone and the watch looses connection,
probably a human will intervene anyway.

2) Fix the intents so that one can use the buttons on the watch. note: not everything seems to be working (yet)

3) Fix build things
a) it seems like gms for wearable is split from main gms? b) fix wear debug version applicationSuffix
c) fix build without mapbox.properties

TESTED=Manually tested that the wear apps seems to mostly function in debug build


wear/src/main/java/org/runnerup/service/ListenerService.java
```diff
@@ -69,6 +70,16 @@ public void onDataChanged(DataEventBuffer dataEvents) {
         }
     }
 
+    @Override
+    public void onPeerConnected(Node peer) {
+        System.err.println("ListenerService.onPeerConnected: " + peer.getId());
```
gerhardol, line 75: only in debug?

gerhardol (commented):
Have not run, but I do not see this affecting use without Wear.

wear/src/main/java/org/runnerup/service/ListenerService.java
jonasoreland: done

jonasoreland:
please take another look ?

gerhardol:
Updated dependencies in https://github.com/jonasoreland/runnerup/pull/1214, fixed a few lints too.
The following should probably be addressed, but that is a little more job.
What Android API on the WearOS device have you been testing with?

Will not be able to test much, just spinning up on a Pixal 8 w Android 15.

--

Lint found errors in the project; aborting build.

Fix the issues identified by lint, or add the issues to the lint baseline via `gradlew updateLintBaseline`.
For more details, see https://developer.android.com/studio/write/lint#snapshot

C:\dev\gc\runnerup\wear\src\main\java\org\runnerup\service\ListenerService.java:136: Error: When targeting Android 13 or higher, posting a permission requires holding the POST_NOTIFICATIONS permission [NotificationPermission]
        notificationManagerCompat.notify(notificationId, notification);
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

   Explanation for issues of type "NotificationPermission":
   When targeting Android 13 and higher, posting permissions requires holding
   the runtime permission android.permission.POST_NOTIFICATIONS.

1 errors, 0 warnings

jonasoreland:
Hi again,

1) I have a 2.45 WearOS Watch
is that new or old ??

2) I'm not sure what you want me to do...
a) should I merge my patch first,
b) should your patch be merged first ?
c) should I fix the lint problem and then ask for new review ?

Thanks!
/Jonas

gerhardol:
> Hi again,
> 
>     1. I have a 2.45 WearOS Watch
>        is that new or old ??
> 
>     2. I'm not sure what you want me to do...
>        a) should I merge my patch first,
>        b) should your patch be merged first ?
>        c) should I fix the lint problem and then ask for new review ?

WearOS uses a certain Android API. 
https://developer.android.com/training/wearables/versions/3/update
WearOS 2 seem to be Android 9, WearOS3 11
WearOS 4 is Android 13, the notifications will likely fail then (not sure what the effect is)

Nothing works before this PR, but this may fail newer watches

jonasoreland:
Fixed the lint problem and some minor tweaks.

Please take another look ?

wear/src/main/AndroidManifest.xml
```diff
@@ -19,6 +19,7 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-feature android:name="android.hardware.type.watch" />
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
```
gerhardol, line 22: Since Android 6 or so, most permissions have to be requested dynamically, the manifest is not enough

jonasoreland: 1) This is on the wear app...how to request that?

2) This fixes the lint issue you pointed too
Maybe merge this and file a follow up bug?

jonasoreland: I installed a the app in an emulated WearOS (latest version)
and indeed, the notification does not show...

But this is not a big deal! One can start the app manually...

Let's merge this...and simply file a bug about that...
ANd if you know, how does not ask for permission on WearOS ??

gerhardol: I have not looked into the specifics for WearOS, but I assume there is no separate APIs, but there may be specific tweaks for Wear that is different to the main app.
This can be a known issue, no blocker.

gerhardol (approved):
Only brief use in phone app, but do not see this to be affecting the normal functionality.

jonasoreland:
Thanks!

PR #1214, opened 2025-05-16
fix: agp 8.7.3

gerhardol:
lint fixes

progress ant+ submodule, no change to the lib

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -86,7 +86,8 @@
                     android:mimeType="application/octet-stream"
                     android:scheme="http"/>
                 <data android:host="*"/>
-                <data android:pathPattern="runnerup.db.export"/>
+                <data android:pathPattern="runnerup.db.export"
+                    tools:ignore="AppLinkUrlError" />
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.VIEW"/>
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 env:
   global:
   - ANDROID_API=34
-  - ANDROID_BUILD_TOOLS=34.0.0
+  - ANDROID_BUILD_TOOLS=36.0.0
   - ADB_INSTALL_TIMEOUT=5
 language: android
 jdk:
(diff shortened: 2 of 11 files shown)


app/AndroidManifest.xml
```diff
@@ -86,7 +86,7 @@
                     android:mimeType="application/octet-stream"
                     android:scheme="http"/>
                 <data android:host="*"/>
-                <data android:pathPattern="runnerup.db.export"/>
+                <data android:pathPattern="/jonasoreland/runnerup/db/export"/>
```
jonasoreland, line 89: ehh...should this be here ??

app/src/wear/org/runnerup/tracker/component/TrackerWear.java
```diff
@@ -476,6 +476,7 @@ public void onMessageReceived(final MessageEvent messageEvent) {
         } else if (Wear.Path.MSG_CMD_WORKOUT_START.contentEquals(messageEvent.getPath())) {
             /* send broadcast to StartActivity */
             Intent startBroadcastIntent = new Intent()
+                    .setPackage(context.getPackageName())
```
jonasoreland, line 479: not sure if this is used elsewhere...but it looks like something that should
be put into function...

jonasoreland (approved):
Looks good to me...
but most of is out of my (current) depth...i dont know what it means :(

app/AndroidManifest.xml
gerhardol: path should be the full path. 
I changed this privately some time ago, will check again

app/src/wear/org/runnerup/tracker/component/TrackerWear.java
gerhardol: this is used at some similar situations

app/AndroidManifest.xml
gerhardol: This is likely an internal filter only, may be removed
Just suppressing for now

gerhardol:
Also updated Java to 17, recommended with agp8

PR #1215, opened 2025-05-16
target android 15

gerhardol:
opt out of edge-to-edge for now
requires insets
This will need to be implemented with Android 16

(first commit is in https://github.com/jonasoreland/runnerup/pull/1214)

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -108,9 +108,11 @@
 
         <activity
             android:name=".view.RunActivity"
-            android:configChanges="keyboardHidden|orientation"/>
+            android:configChanges="keyboardHidden|orientation"
+            android:theme="@style/AppTheme"/>
 
-        <activity android:name=".view.SettingsActivity"/>
+        <activity android:name=".view.SettingsActivity"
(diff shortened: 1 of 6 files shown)


app/AndroidManifest.xml
```diff
@@ -86,7 +86,7 @@
                     android:mimeType="application/octet-stream"
                     android:scheme="http"/>
                 <data android:host="*"/>
-                <data android:pathPattern="runnerup.db.export"/>
+                <data android:pathPattern="/jonasoreland/runnerup/db/export"/>
```
jonasoreland, line 89: looks weird ?

jonasoreland (approved):
looks good to me
(with question about usage of my name in a path??)

app/AndroidManifest.xml
gerhardol: this is part of #1214

gerhardol:
There are some visual issues in Hr settings, it seems like the style is not applied there.
Will set as draft.

gerhardol:
fixed the remaining activities, rebased on #1214

PR #1221, opened 2025-05-27
chore: prepare for 2.9.0.1

gerhardol:
add Wear version with one compared to phone app

enough to publish Wear?


jonasoreland (commented):
As i read the instructions

1) the version code should be *the same*
2) the wear app must be in the same app bundle
i.e. not uploaded separately...

and...we should remove all the applicationVariants stuff from */build.gradle
that should all be done with app bundle...

wdyt ?

gerhardol:
https://developer.android.com/training/wearables/packaging

Version can be independent (so I will revert the +1), package name must be the same (it is), formfactor to be added to Play Console. then a separate bundle is uploaded.
Can you check the common reasons that Wear Apps are rejected first?

jonasoreland:
1) I checked https://developer.android.com/training/wearables/packaging#reasons-for-rejection
- it works now (but didn't) work for round
- we must add "Wear OS" to the store listing
- we must add wear screenshots

Other than that I think we're fine.

2) re separate bundle: https://developer.android.com/training/wearables/packaging#planning-for-play-store
"If you have a phone APK in addition to a watch APK, you must use the [Multi-APK delivery method](https://developer.android.com/google/play/publishing/multiple-apks)."

I think that means that it should be 1 bundle...not a separate one...
But I'm not an expert...and I don't remember how this is done at all...

jonasoreland:
Update: I logged into "https://play.google.com/console" but did not find where to modify store listing
or screenshots... :(

gerhardol:
Console: Test&Release, Setup, Advanced settings, Form factors
I added Wear as a form factor, follow the first item to add screen shots and to update description (last line at least). Normal screenshots should be updated too.
Then we release new bundles.

jonasoreland:
yay, updated 3 wear screenshots.

PR #1222, opened 2025-05-29
Wear SmallRound

jonasoreland:
Adjust layouts to work better on SmallRound
(I used to have a rectangular wear...not I have a small round)


gerhardol (approved):
Have not run or really tried to break down the changes, but see nothing strange.

I want square watches back, but will not go with Apple.

PR #1231, opened 2025-06-08
More fixes for Wear OS (based on play store feedback)

jonasoreland:
1) Remove the "action bar" that was shown when
the app started...and that was incorrectly centered (how to fix that??). This was a complaint from play console. This is fixed in manifest + styles.

2) Add the OngoingActivity concept to the Notification. This does nothing on my old watch, but seems to work on never versions (tried in emulator)


jonasoreland:
Damn, this works fine in emulator...but break my fossil-5e...
I'll update this pr once I get both emulator...and real device to work.

jonasoreland:
ok, now it seems to work both on my watch...and the emulator...
I also (partially) switched to new apis, not deprecated (yet?), yay!

gerhardol (approved):
Brief review, cannot test.
Merge to prepare for Play update

PR #1232, opened 2025-06-09
fix: edge-to-edge insets

roberi:
This pull request addresses an oversight from #1226 where insets were not applied to the following activities, causing UI elements to overlap with system bars.

- OAuth2Activity
- ManualActivity

This PR also fixes an issue when switching/selecting between different audio cues in AudioCueSettingsFragment. The new fragment's UI got obscured by the AppBar (due to lost insets?).

Tested in Android Studio on AVDs running Android 11 and Android 15.

### DIFF
--- a/app/src/main/org/runnerup/view/AudioCueSettingsFragment.java
+++ b/app/src/main/org/runnerup/view/AudioCueSettingsFragment.java
@@ -21,10 +21,6 @@ import android.widget.EditText;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
-import androidx.core.graphics.Insets;
-import androidx.core.view.OnApplyWindowInsetsListener;
-import androidx.core.view.ViewCompat;
-import androidx.core.view.WindowInsetsCompat;
 import androidx.preference.Preference;
 import androidx.preference.Preference.OnPreferenceClickListener;
 import androidx.preference.PreferenceFragmentCompat;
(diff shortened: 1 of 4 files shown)


roberi (commented):
Note, ViewUtil.Insets is not utilized (the commit message is wrong). I simply moved the OnApplyWindowInsetsListener from the AudioCueSettingsFragment to the AudioCueSettingsActivity.

app/src/main/org/runnerup/view/AudioCueSettingsFragment.java
```diff
@@ -154,19 +150,6 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
       }
       spinner.setOnSetValueListener(onSetValueListener);
     }
-
-    ViewCompat.setOnApplyWindowInsetsListener(
```
gerhardol, line 158: For me this works, I believe. What is it that you see?
(Can change this, mostly curious.)

gerhardol (approved):
OK to squash changes?

app/src/main/org/runnerup/view/ManualActivity.java
```diff
@@ -74,6 +75,8 @@ public void onCreate(Bundle savedInstanceState) {
     manualPace = findViewById(R.id.manual_pace);
     manualPace.setVisibility(View.GONE);
     manualNotes = findViewById(R.id.manual_notes);
+
+    ViewUtil.Insets(findViewById(R.id.tab_manual), true);
```
gerhardol, line 79: sidenote: The divider in the bottom should be removed to use edge-to-edge properly
There are other situations the screen could be used better, but there are many other UI tweaks that should be done

app/src/main/org/runnerup/view/AudioCueSettingsFragment.java
roberi: It looks like this for me, both in an AVD and on my Pixel 9. The left image is when the activity is first started. The right image is after selecting a different audio cue in the spinner.

image

roberi:
> OK to squash changes?

Yes.

gerhardol:
Thanks!

Issue #1237, opened 2025-06-19
RunnerUp Wear is rejected by play store


jonasoreland:
1. Issue found: Missing ongoing activity
When a user has an ongoing activity, your app did not do one or more of the following:

Show the ongoing activity indicator on the watch face.
Update recent apps with the appropriate app launcher chip for the ongoing activity.
Reference the ongoing activity from the tile if the tile is present in the tile carousel. For more information, see [Ongoing Activity](https://c.gle/ANiao5qQ8WV8FIYBKQLjPO-xyvmX08IXzDtiNCPsXZ0mlsEYqwUYPPskAWX8OQiStbgpaikWl4UXYe48PXwtKxQSQcw77N29D4BQ9tg-pm1jtn7WcwMIFg_2GV5ySNMrQyVzy6vS1U3M7gn3VgdNr7zXhAR9xt7FOjERdGK5LD5SMGlYzcUDKkcwTlChAUS5ggCgBQTCNc0X4tMTypQNO0wprzMe_M6Wm15Evzur-cSbsGoQeLADAz_3oR0o9oqRvINtNwH9mXH8bZ1VTae54JIkdSOf6P1Hcl98Du65lwAwxvotW9xLmtVes0OXb1bigD6U).
Issue details

We found an issue in the following area(s):

Version code 21000353
About the [Wear App Quality Guidelines](https://c.gle/ANiao5oLFc0S3WQArEILxwgxxooEHwc9KRcUE7Y0qkCcLbk30YGscf8wRheqBLT9Dj1yBy7i98l-5ic4rYAfjIjslwBIPdu0jH7fBbfgp9kA-2xDWShI7LwT7RoV_n5XURINz-986FSysYj3JZ3_TFwh07E8rqtSyOjsb-No4POZ2wrlkw)
We do not accept apps into Wear OS that don't adhere to [Wear App Quality](https://c.gle/ANiao5oLFc0S3WQArEILxwgxxooEHwc9KRcUE7Y0qkCcLbk30YGscf8wRheqBLT9Dj1yBy7i98l-5ic4rYAfjIjslwBIPdu0jH7fBbfgp9kA-2xDWShI7LwT7RoV_n5XURINz-986FSysYj3JZ3_TFwh07E8rqtSyOjsb-No4POZ2wrlkw) Guidelines.

2. Issue found: Watch shapes
Your app or watch face content does not meet one or more of the following visual quality requirements:

Fits within the physical display area.
[...]

PR #1233, opened 2025-06-15
Replace deprecated TabActivity in MainLayout

roberi:
(Sorry for a big pull request. I couldn't think of a good way to split it into multiple PRs.)

The `MainLayout` currently uses `TabActivity`, `TabHost`, and `TabWidget` for its tabbed navigation. These have all been deprecated for a long time. This PR replaces them with AndroidX components.

Key changes:

- `MainLayout` now extends `AppCompatActivity` instead of `TabActivity`.
- Tab navigation and content display are implemented using `ViewPager2` and `TabLayout`.
- A `FragmentStateAdapter` (`BottomNavFragmentStateAdapter`) manages fragment lifecycles within `ViewPager2`.
- `TabLayout` is synchronized with `ViewPager2` via `TabLayoutMediator`.
- The layout `R.layout.main` has been updated to reflect these changes.
- A `SettingsContainerFragment` is added to host the app's settings hierarchy (e.g., `SettingsFragment` and its sub-settings fragments).
- Swipe gesture on `ViewPager2` is disabled to maintain the original click-to-navigate behavior. 
- Smooth scrolling in `TabLayoutMediator` is disabled to mimic previous behavior.

I have manually tested in Android Studio on AVDs using API levels 30 and 35. More testing on real devices is probably needed.

Remaining Work/Considerations:

1. Replace usage of deprecated `TabHost`/`TabWidget` in `StartFragment` and `DetailActivity` with `ViewPager2`/`TabLayout`.
2. In AndroidManifest, the following activities have their `android:parentActivityName` attribute changed to `MainLayout`, but should probably have a different parent for **Up navigation**.
   - `DetailActivity` (had `HistoryActivity`)
   - `ManualActivity` (had `HistoryActivity`)
   - `HRSettingsActivity` (had `SettingsActivity`)
   - `HRZonesActivity` (had `SettingsActivity`)
3. Enable swiping and/or smooth scrolling between bottom tabs?

Related issues:

* #648
  * Can it be closed?
* #1210
  * The [same test](https://github.com/jonasoreland/runnerup/issues/1210#issuecomment-2831394824) no longer indicates a resource leak. @zh222, it would be great if you could test again.

PR #1239, opened 2025-06-22
Wear cleanup

jonasoreland:
This PR does the following
1) create a utility class WearNotifier that wakes up wear app when
StartActivity starts. This is an attempt to fix https://github.com/jonasoreland/runnerup/issues/1237 (part 1)

2) create a WearableClient that is shared between app and wear
that encapsulates the google wear api...and migrates read/put data

Goal is to put also messages and listeners in there.

### DIFF
--- a/app/src/main/org/runnerup/view/StartActivity.java
+++ b/app/src/main/org/runnerup/view/StartActivity.java
@@ -131,6 +131,7 @@ public class StartActivity extends AppCompatActivity implements TickListener, Gp
 
   private View wearOsIndicator = null;
   private TextView wearOsMessage = null;
+  private TrackerWear.WearNotifier mWearNotifier = null;
 
   boolean batteryLevelMessageShown = false;
   private Boolean exit = false;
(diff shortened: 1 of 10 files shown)


gerhardol:
The apis are updated, looks OK to me in a brief check.
Have not run.

jonasoreland:
still struggling a bit with github work-flow: is your comment the same as "approve" / "ok to merge" ?

gerhardol:
> still struggling a bit with github work-flow: is your comment the same as "approve" / "ok to merge" ?

It was a review (not formally set as such) but not approval: I would like to take a second look as I was tired yesterday.
(But will likely not have more opinions and do not oppose merging.)

jonasoreland:
ok!

app/src/main/org/runnerup/view/StartActivity.java
```diff
@@ -327,6 +328,9 @@ public void handleOnBackPressed() {
                 }
               }
             });
+
+    mWearNotifier = new TrackerWear.WearNotifier(getApplicationContext());
+    mWearNotifier.onCreate();
```
gerhardol, line 333: Changed with fragments in #1233 
```suggestion
    mWearNotifier.onViewCreated();

```

gerhardol (approved):

PR #1241, opened 2025-06-29
Start of activities without GPS

jonasoreland:
This patch contains nop of refactorings as preparation for activities without gps.

The "main change" is adding early out in lots of places,
i.e. from

if (condition) {
  long blob of code
} else {
  return;
}

to 

if (!condition) {
  return;
}
long blob of code;


app/src/main/org/runnerup/view/StartActivity.java
```diff
@@ -867,27 +871,83 @@ private void toggleStatusDetails() {
...
+      if (mGpsStatus.isEnabled()) {
+        gpsEnable.setText(org.runnerup.common.R.string.Start_GPS);
+      } else {
+        gpsEnable.setText(org.runnerup.common.R.string.Enable_GPS);
+      }
       gpsEnable.setVisibility(View.VISIBLE);
+      return;
+    } while (false);
+
+    gpsEnable.setVisibility(View.GONE);
```
gerhardol, line 946:
```suggestion
    if (mGpsStatus.isStarted()) {
        gpsEnable.setVisibility(View.GONE);
    } else {
        gpsEnable.setVisibility(View.VISIBLE);
        int textResId = mGpsStatus.isEnabled() ?
                        org.runnerup.common.R.string.Start_GPS :
                        org.runnerup.common.R.string.Enable_GPS;
        gpsEnable.setText(textResId);
    }
```

app/src/main/org/runnerup/view/StartActivity.java
```diff
@@ -867,27 +871,83 @@ private void toggleStatusDetails() {
...
+      if (Objects.requireNonNull(tabHost.getCurrentTabTag()).contentEquals(TAB_ADVANCED)
+          && advancedWorkout == null) {
+        break;
+      }
+
+      startButton.setVisibility(View.VISIBLE);
+      return;
+    } while (false);
+
+    startButton.setVisibility(View.GONE);
```
gerhardol, line 927:
```suggestion
  // Check conditions for showing the start button
  boolean showStartButton = mGpsStatus.isStarted() &&
                            mGpsStatus.isLogging() &&
                            mGpsStatus.isFixed() &&
                            mTracker != null &&
                            mIsBound &&
                            !(Objects.requireNonNull(tabHost.getCurrentTabTag()).contentEquals(TAB_ADVANCED) && advancedWorkout == null);

  // Set the visibility of the start button based on the conditions
  startButton.setVisibility(showStartButton ? View.VISIBLE : View.GONE);

```

PR #1248, opened 2025-07-18
Fix problem occuring with really short intervals (e.g 15s)

jonasoreland:
In that case the audio cues would "add up" so that one missed the start/stop and count downs.

This patch introduces a PRIO so that higher prio interrupt lower prio.

This fixes immediate problem...maybe we should avoid adding several "speak" commands in total and do our own scheduling, using some sort of key per cue...todo...

Tested/iterated a couple of times with short intervals, and tested with soft runs (no change)

### DIFF
--- a/app/src/main/org/runnerup/workout/feedback/AudioFeedback.java
+++ b/app/src/main/org/runnerup/workout/feedback/AudioFeedback.java
@@ -18,7 +18,6 @@
 package org.runnerup.workout.feedback;
 
 import android.content.Context;
-import android.speech.tts.TextToSpeech;
 import android.widget.Toast;
 import java.util.HashMap;
 import org.runnerup.BuildConfig;
@@ -40,6 +39,11 @@ public class AudioFeedback extends Feedback {
   private Intensity intensity;
   RUTextToSpeech textToSpeech;
   Formatter formatter;
+
+  // Higher prio will interrupt (flush) lower prio.
+  public static final int PRIO_CUE = 0;
+  public static final int PRIO_COACH = 1;
+  public static final int PRIO_COUNTDOWN = 2;
 
   public AudioFeedback(int msgId) {
     super();
@@ -121,7 +125,7 @@ public class AudioFeedback extends Feedback {
       if (BuildConfig.DEBUG) {
         Toast.makeText(ctx, msg, Toast.LENGTH_SHORT).show();
       }
-      textToSpeech.speak(msg, TextToSpeech.QUEUE_ADD, null);
+      textToSpeech.speak(msg, PRIO_CUE, /* flush= */false, null);
     }
   }
 }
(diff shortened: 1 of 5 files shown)


app/src/main/org/runnerup/workout/feedback/AudioFeedback.java
```diff
@@ -1,127 +1,131 @@
...
+  private String msgTxt = null;
+  private final Event event = null;
+  final Scope scope;
+  final Dimension dimension;
+  private Intensity intensity;
+  RUTextToSpeech textToSpeech;
+  Formatter formatter;
+
+  // Higher prio will interrupt (flush) lower prio.
+  public static final int PRIO_CUE = 0;
```
gerhardol, line 44: make this an enum?

app/src/main/org/runnerup/workout/feedback/RUTextToSpeech.java
```diff
@@ -91,92 +100,58 @@ public Boolean isAvailable() {
...
-  int speak(String text, int queueMode, HashMap<String, String> params) {
+  int speak(String text, int prio, boolean flush, HashMap<String, String> params) {
     if (!isAvailable()) {
       return 0;
     }
 
-    final boolean trace = true;
-    if (queueMode == TextToSpeech.QUEUE_FLUSH) {
-      // Unused in RU
+    boolean trace = true;
```
gerhardol, line 108: why remove final?

app/src/main/org/runnerup/workout/feedback/AudioFeedback.java
```diff
@@ -1,127 +1,131 @@
-/*
```
gerhardol, line 1: change back to \r\n to limit noise
we could change all files to \n
(I expected the formatter to set this...)

gerhardol (commented):
limited review, probably OK

app/src/main/org/runnerup/workout/feedback/AudioFeedback.java
jonasoreland: why wait ?
I'll revert here, but do an additional commit.
so that change can more easily been seen...

app/src/main/org/runnerup/workout/feedback/RUTextToSpeech.java
jonasoreland: misstake :)

app/src/main/org/runnerup/workout/feedback/AudioFeedback.java
jonasoreland: done

jonasoreland:
PTAL

app/src/main/org/runnerup/workout/feedback/AudioFeedback.java
gerhardol: Just to change all in one separate commit.
It is then easier to ignore non functional changes in diff and ignore. 
git-blame can automate this: https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-filefile, for instance setting blame.ignoreRevsFile  to .git-blame-ignore-revs
(I use that in some work projects, but it is convenient to configure blame this way also if there is no  .git-blame-ignore-revs).

PR #1250, opened 2025-07-21
run dos2unix on all java files

jonasoreland:


gerhardol:
Also converted a few other files I suggest that are included

gerhardol (approved):

PR #1251, opened 2025-07-21
Add new sport 'Treadmill'

jonasoreland:
It is designated as without gps...
but no functionality is (yet) connected to this... coming in subsequent patches.


common/src/main/res/values/array.xml
```diff
@@ -57,6 +57,7 @@
     <item>Other</item>
     <item>Orienteering</item>
     <item>Walking</item>
+    <item>Treadmill</item>
```
gerhardol, line 60: This adds Treadmill to English only. All other languages will not have this entry.
I believe it kind of works since this is added in the end, just not visible.

I rather have a "GPS" checkbox, then also indoor cycling is supported, as well as gym.

gerhardol (commented):
-1
I rather have a "GPS" checkbox, then also indoor cycling is supported, as well as gym.

jonasoreland:
> -1 I rather have a "GPS" checkbox, then also indoor cycling is supported, as well as gym.

1) Gym is 100% a new sport...(i.e. it is not other without gps)

2) I think treadmill is easier to understand/use than running w/o gps, less clutter in UI
b) We should also add the feature to manually enter distance for treadmill

3) However if someone (you?) adds a good UI for "no gps checkbox"...then I "don't mind"
b)If that feature is added, making so that the checkbox is invisible for treadmill should be easy ?

4) Adding stationary bike as separate sport will also be trivial...

---

Summary:
5) I think this is better
6) Have you checked any competitors ?

gerhardol:
Suggestion: Keep "Start GPS" button in Start enabled all the time, with "Stop GPS" if GPS is started (you can use Back now if GPS is not autostarted, to be removed then.
This part is not really needed to get started on recording without GPS.

Always add Start (start_fab.xml) maybe with other colors if no GPS.

gerhardol:
In the long run, I suggest that only Advanced workouts are supported. Basically current sports added as a workout.
Each workout has a field with the current "sports" are added as a field, really just tcx categories for upload (Run/Bike/Other), with some heuristics for e.g. Strava from the name.

jonasoreland:
> Suggestion: Keep "Start GPS" button in Start enabled all the time, with "Stop GPS" if GPS is started (you can use Back now if GPS is not autostarted, to be removed then. This part is not really needed to get started on recording without GPS.
> 
> Always add Start (start_fab.xml) maybe with other colors if no GPS.

agree, this is my plan

jonasoreland:
> In the long run, I suggest that only Advanced workouts are supported. Basically current sports added as a workout. Each workout has a field with the current "sports" are added as a field, really just tcx categories for upload (Run/Bike/Other), with some heuristics for e.g. Strava from the name.

don't really understand what you mean...but it also doesn't feel strictly related to this pr ?

jonasoreland:
submit as is for now ?
or maybe I should add the treadmill to all languages...
or something else :)

wdyt ?

gerhardol:
I added #1253 as an alternative

> don't really understand what you mean...but it also doesn't feel strictly related to this pr ?

I just feel adding this sport contradicted to a possible UI cleanup, where users can add whatever sport they want
Much simpler

jonasoreland:
> I added #1253 as an alternative
> 
> > don't really understand what you mean...but it also doesn't feel strictly related to this pr ?
> 
> I just feel adding this sport contradicted to a possible UI cleanup, where users can add whatever sport they want Much simpler

ahh...then I understand,

but I disagree
Having free-text form rather than drop down seems like a bad idea to me...
Imo it is not a cleanup, "simpler" for us => but will just make it harder for user.

Also, my treadmill suggestion will attach behavior to the sport (manual adding of distance).

---

I'm in the camp where the "easy things should be easy"...

- which I also think that merging Interval and Advance is a bad idea...

- which is also infact a reason that I'm not super happy about the current structure of the preferences...
I think the settings that user are most likely to change should be on "first page" and various others
should be hidden (but available) in "advanced" sub-sections---

---

Again, I propose we merge this now...
and (maybe) continue arguing in parallel...

jonasoreland:
other cas where "stop gps" does not really make sense
i.e. properties of sport is imo better

- gym: why should I stop GPS...and obviously I should not add distance afterwards ?

gerhardol:
Of course settings like GPS, autopause ond stepsensor need  to be added to the workout.
The current sports should be in the workout list as well as run intervalls.
So just simpler.
Arrays are more or less immutable, they are hard to use with translations. This should push translations and fetch tem (so all translation will need to be updated) otherwise the feature is only usable in English.
(Note that there are a few manual overrides, empty language files are removed as well as nl-EN).
This PR is kindof a blocker for releases from master, until the other functionality is there. I opened #1254 to be merged before, to prepare f-droid at least.

But I do not block the change, just voting against it. You have regardless the majority vote....
Still prefer to just toggle GPS right now.

jonasoreland:
One more round of comments/questions/arguments :) (maybe last?)

1) I added the string to the other arrays
b) don't remember how translation works! (not even sure it was me adding it??)

2) release blocker: I think it could be fine with just adding the choice, add then "the rest" of the feature.
I.e. not blocking release.
but I'm also fine with waiting to merge this part...and complete the rest...it is not that much remaining.

wdyt ?

3) #1254 obviously lgtm,

---

gerhardol:
>     1. I added the string to the other arrays
>        b) don't remember how translation works! (not even sure it was me adding it??)

The English source need to be pushed to Transifex, then pulled, look if there are overrides (I may have retained some changes in some cleanup), then fixup and push translations, then pull.
If this is the way it should be done, Trainer (or Indoor running, then Treadmill should be Indoor running, also including other indoor) should be added now, it is tedious to do this part again.

>     2. release blocker: I think it could be fine with just adding the choice, add then "the rest" of the feature.
>        I.e. not blocking release.
>        but I'm also fine with waiting to merge this part...and complete the rest...it is not that much remaining.

This is not my preferred solution, but I am fine with being voted down. It is life.
But if the rest is close, then I suggest you base next PR on top of this PR and both are merged in serial.

>     3. [fix: prepare 2.9.0.6 #1254](https://github.com/jonasoreland/runnerup/pull/1254) obviously lgtm,

Have not decided if to move the button or not. If the Start button is always added (so you can press start before GPS locks too), then current version is OK, otherwise it should be moved.
I want this regardless.

jonasoreland:
> The English source need to be pushed to Transifex, then pulled, look if there are overrides (I may have retained some changes in some cleanup), then fixup and push translations, then pull. If this is the way it should be done, Trainer (or Indoor running, then Treadmill should be Indoor running, also including other indoor) should be added now, it is tedious to do this part again.

ehh...didn't really follow :(
I got that it is tedious...

a) and that you want to add several sports at once ? Which sports ?
I think 'Treadmill' is good by itself...

b) shall I be doing the transifex push/pull tango you describe ?
or is submitting one/several patches only translated to english (like I propose) first
and then the tango is done ?

> 
> > ```
> > 2. release blocker: I think it could be fine with just adding the choice, add then "the rest" of the feature.
> >    I.e. not blocking release.
> >    but I'm also fine with waiting to merge this part...and complete the rest...it is not that much remaining.
> > ```
> 
> This is not my preferred solution, but I am fine with being voted down. It is life. But if the rest is close, then I suggest you base next PR on top of this PR and both are merged in serial.

ok, that is fine too...
 
> > ```
> > 3. [fix: prepare 2.9.0.6 #1254](https://github.com/jonasoreland/runnerup/pull/1254) obviously lgtm,
> > ```
> 
> Have not decided if to move the button or not. If the Start button is always added (so you can press start before GPS locks too), then current version is OK, otherwise it should be moved. I want this regardless.

ack...

jonasoreland:
hmm....maybe I should add all the features in one branch...
(gpsless and manual distance)...

and then add new sports (taking translations into account) in a separate commit once the features are already there...
wdyt ?

gerhardol:
> hmm....maybe I should add all the features in one branch... (gpsless and manual distance)...
> 
> and then add new sports (taking translations into account) in a separate commit once the features are already there... wdyt ?

yes, as long as you keep commit feature separated, I can somehow review.

> ehh...didn't really follow :(
I got that it is tedious...

Using arrays are complicated for translations.
It is possible to do a hybrid where translated files are not pushed back to Transifex so the translators can see that there are missing strings.
I have cleaned up this once already, but then some languages were out of order too...
It is easier if this manual fixes are done only once, so if you are sure that it is this way you want to record non-GPS (and not use a simple GPS off switch as in #1253) other sports should be added now.
(If I fix it, I prefer only once.)

jonasoreland:
Thanks for explanation about arrays& translations...i had no idea.

1) I'll explore getting rid of the array for sport,
It feels not great to have to think about such 'technicalities'
https://github.com/jonasoreland/runnerup/pull/1257

2) i still think new sports with new properties is a good addition

b) I'll continue on treadmill, trying to split into digestable pieces...it
is almost work now in my sandbox...


3) i'm not 100% sure what the turn of gps pr is supposed to do...(but
haven't looked that carefully to be honest...the code looks great, but i
don't understand the use case...)

jonasoreland:
FYI: this branch now has everything i think is needed to merge. I made my first attempt with treadmill this morning...it worked :)


I plan to split the full thing into 3 separate pull-requests (branches).
1) manual distance
2) no gps
3) treadmill (enabling the 2 above)

Then there will probably be a few follow ups/bug fixes:
4) sport without gps can filter away audio cues for distance/pace/etc
5) the graph which should be showing my heart rate over time doesn't
And similar bits and pieces...

Wdyt?

/Jonas

PR #1253, opened 2025-07-23
feat: disable GPS button in start

gerhardol:
Alternative to #1251


gerhardol:
will not use this way

PR #1257, opened 2025-07-29
Deprecate sportEntries (string-array) in favor of static map in Sport

jonasoreland:
This patchs adds a map between DB.ACTIVITY.SPORT_xxx and R.strings.SPORT_xxx that is intended to be used instead of array.sport_entries, which apparently is hard to handle wrt to translations.

The patch currently use the existing sport_entries as bases so that languages that has not yet added R.strings.SPORT_xxx will be unaffected.

### DIFF
--- a/app/src/main/org/runnerup/workout/Sport.java
+++ b/app/src/main/org/runnerup/workout/Sport.java
@@ -18,6 +18,9 @@
 package org.runnerup.workout;
 
 import android.content.res.Resources;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 import org.runnerup.R;
 import org.runnerup.common.util.Constants.DB;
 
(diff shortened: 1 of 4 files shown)


jonasoreland:
I think this will the problems with adding new sport entries...

app/src/main/org/runnerup/workout/Sport.java
```diff
@@ -38,14 +41,54 @@ public int getDbValue() {
...
+    result.put(DB.ACTIVITY.SPORT_ORIENTEERING, org.runnerup.common.R.string.SPORT_ORIENTEERING);
+    result.put(DB.ACTIVITY.SPORT_WALKING, org.runnerup.common.R.string.SPORT_WALKING);
+    return Collections.unmodifiableMap(result);
+  }
+  private static final Map<Integer,Integer> gSportToStringMap = createSportToStringMap();
+
+  public static String[] getStringArray(Resources res) {
+    String[] ret = new String[DB.ACTIVITY.SPORT_MAX + 1];
+
+    {  // Backward compability. TODO: Remove once new sport strings are translated.
```
gerhardol, line 58: Looks good.
I was to suggest an alternative where the new strings are used if they are defined, the array otherwise, but my PC now fail to start...

gerhardol (approved):

PR #1261, opened 2025-08-01
feat: Workout without GPS

jonasoreland:
This patch adds capability of doing a workout without gps, which is mainly 2 parts
1) juggle state in StartFragment
- toggle from gps-on to gps-off is instant
- toggle from gps-off to gps-on requires resetting Tracker since it does not support moving 'backwards' on state-machine.
- fix the views
2) Fix the actual running w/o gps (TrackerGPS)
- always use a new location
- add gps off toggle

The feature is currently unused but is planned to be used with Treadmill (and stationary bike?)


gerhardol:
How are export working?
I assumed that all formats would be affected.

jonasoreland:
no format affected as far as I can tell.
but the tcx/gpx track will have identical gps position for all datapoints => works fine.

jonasoreland:
hmm...retested => found track without data points...that "works" but is maybe not what one would expect,
maybe this is also why i didn't get a HR graph..

it doesn't crash...and the tcx/gpx is "sort of correct"
ok if I fix in follow up ? (there will be a few!)

jonasoreland:
Fixed export of gpx/tcx.
PTAL

gerhardol:
> hmm...retested => found track without data points...that "works" but is maybe not what one would expect, maybe this is also why i didn't get a HR graph..

There are other export formats other than tcx/gpx and sure some other situations.

> it doesn't crash...and the tcx/gpx is "sort of correct" ok if I fix in follow up ? (there will be a few!)

I have not really reviewed the latest PRs yet, so just quick comments.
It can be hard to keep the changes as 'advanced', master seem to be unstable with these changes.
But we could make releases from a release branch if needed.

Related: Should we drop Wear in Play for now and try again later?
That might be required to get 2.9 approved. (No major changes though.)

jonasoreland:
On Fri, Aug 1, 2025, 10:39 Gerhard Olsson ***@***.***> wrote:

> *gerhardol* left a comment (jonasoreland/runnerup#1261)
> <https://github.com/jonasoreland/runnerup/pull/1261#issuecomment-3143738130>
>
> hmm...retested => found track without data points...that "works" but is
> maybe not what one would expect, maybe this is also why i didn't get a HR
> graph..
>
> There are other export formats other than tcx/gpx and sure some other
> situations.
>
> it doesn't crash...and the tcx/gpx is "sort of correct" ok if I fix in
> follow up ? (there will be a few!)
>
> I have not really reviewed the latest PRs yet, so just quick comments.
> It can be hard to keep the changes as 'advanced', master seem to be
> unstable with these changes.
>
I don't understand what you mean. There is to my knowledge nothing
'unstable' ?? Can you clarify?

This pr...and the next are (should be) NOP from user perspective...until
the final piece is merged (i.e. treadmill).


But we could make releases from a release branch if needed.
>

I think a branch is pointless...we have no testing that i know of...and
probably only me would use the branch...

> Related: Should we drop Wear in Play for now and try again later?
> That might be required to get 2.9 approved. (No major changes though.)
>
1) can i also try calling them once?
I'll try before end of day, or first thing Monday.
If that doesn't give anything,
- let's disable, make release, start argument with them again

2) regardless, regarding treadmill prs
I'm in the camp of merging unless
a) it.is.obviously broken or half baked
b) it will introduce regression for an *existing* use case...

I don't think neither a or b or true => i think we should merge...

A compromise would be to make a 2.9 release and then merge...(which is fine
for me)

Wdyt?

/Jonas

gerhardol:
> I don't understand what you mean. There is to my knowledge nothing
'unstable' ?? Can you clarify?

With 'unstable' I mean features that do not fully work, without clearly marking them as 'experimental'.
I believe some parts of this is here. Good if I am wrong.

It is nice if master is always ready to release, but longterm feature branches adds overhead.

The major reason for a release is to sort out Android 15, so please try to contact Play store again.

jonasoreland:
> > I don't understand what you mean. There is to my knowledge nothing
> > 'unstable' ?? Can you clarify?
> 
> With 'unstable' I mean features that do not fully work, without clearly marking them as 'experimental'. I believe some parts of this is here. Good if I am wrong.

i think you are wrong
but please add 'actual' comments on my 2 pull-requests...not only speculation :)

> It is nice if master is always ready to release, but longterm feature branches adds overhead.

re always ready to release: yes, that is obvious

> 
> The major reason for a release is to sort out Android 15, so please try to contact Play store again.

i'll do it first thing monday morning...(added to calendar!)

if it doesn't work, let's do 2.9 w/o wear...
also, what is needed for android 15 ? is there a bug and/or decription ? I have no idea

/Jonas

roberi:
What is the issue with wear? Still the "Watch shapes" in #1237?

jonasoreland:
> What is the issue with wear? Still the "Watch shapes" in #1237?

Unknown!
See https://support.google.com/googleplay/android-developer/thread/358380132/my-wear-app-is-rejected-incorrectly-and-without-proper-explanation

roberi:
> > What is the issue with wear? Still the "Watch shapes" in #1237?
> 
> Unknown! See ...

Thanks for the info. I hope the issue gets resolved soon (as they seem to have accepted your appeal).

jonasoreland:
PTAL ?

PR #1263, opened 2025-08-10
Add 2 new properties for a Sport

jonasoreland:
- Manual distance A user is prompted to enter distance at end of workout

- WithoutGps The GPS location is not saved. No distance/speed/pace is computed


jonasoreland:
@roberi : Care to review ?

PR #1264, opened 2025-08-11
Remove detail_nomap.xml in favor of handling in DetailActivity.java

jonasoreland:
This is a nop refactoring.

Move handling on no map into java-code rather than using different layouts...I think that is easier.

- Rename directory "nomapbox" to "osmdroid".
- Add a directory "nomap".
- Add build option for running w/o map, previously the code supported it... but I found no way of actually building it?


app/build.gradle
```diff
@@ -27,10 +27,12 @@ android {
...
                 java.srcDirs += ['src/nonwear']
             }
-            if (rootProject.ext.useMapBox) {
+            if (rootProject.ext.noMap) {
+                java.srcDirs += ['src/nomap']
+	    } else if (rootProject.ext.useMapBox) {
                 java.srcDirs += ['src/mapbox']
-            } else {
-                java.srcDirs += ['src/nonmapbox']
+	    } else {
```
gerhardol, line 34: tabs were used

```suggestion
            } else if (rootProject.ext.useMapBox) {
                java.srcDirs += ['src/mapbox']
            } else {
```

gerhardol (commented):
+1
Seem fine but have not run. Cannot think of anything alse right now.

PR #1266, opened 2025-08-12
Request POST_NOTIFICATIONS permission on Wear OS

roberi:
Starting in Wear OS 4 (Android 13, API level 33), ongoing notifications are affected by the notification runtime permission. In the Wear app, there is currently no functionality to request this permission. If this permission is required and not granted, the ListenerService.showNotification() method exits. 

```java
// check if we have permission to post notification
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU
    && checkSelfPermission(android.Manifest.permission.POST_NOTIFICATIONS)
        != android.content.pm.PackageManager.PERMISSION_GRANTED) {
  return;
}
```
This PR introduces both a generic way to handle runtime permission in the  Wearapp and specifically addresses the POST_NOTIFICATIONS permission requirement for Android 13 (API 33) and higher.

Key Changes include:

- A new reusable component RequestPermissionActivity to manage requests for any Android runtime permission (for now only the POST_NOTIFICATIONS permission, but later also BODY_SENSORS for heart rate monitoring).
-  If needed, this activity displays a rationale to the user before the system permission dialog appears.
-  It reports the outcome of the permission request back to the calling component (if they are interested).
-  If the POST_NOTIFICATIONS permission is not already granted, the MainActivity (in the wear module) utilizes the new RequestPermissionActivity to prompt the user.

https://github.com/user-attachments/assets/453b5090-55f7-4894-a9c6-429c0a5c21e4

### DIFF
--- a/wear/src/main/java/org/runnerup/view/RequestPermissionActivity.java
+++ b/wear/src/main/java/org/runnerup/view/RequestPermissionActivity.java
@@ -78,15 +78,10 @@ public class RequestPermissionActivity extends AppCompatActivity {
           sendPermissionResultAndFinish(true);
         } else {
           Log.w(TAG, permissionToRequest + " denied by user.");
-          if (shouldShowRequestPermissionRationale(permissionToRequest)) {
-            Log.d(TAG, "resultCallback: Rationale needed. Making UI visible.");
-            rationaleUIRoot.setVisibility(View.VISIBLE);
-          } else {
-            Log.d(
-                TAG,
-                "resultCallback: " + permissionToRequest + " denied and policy prevents asking.");
-            sendPermissionResultAndFinish(false);
-          }
+          // TODO: As per Android guidelines, implement UI to inform the user that the specific
+          //  feature (requiring permissionToRequest) is unavailable due to denial.
+          // For now, just send the denial result back to the caller.
+          sendPermissionResultAndFinish(false);
         }
       };
 
@@ -147,6 +142,9 @@ public class RequestPermissionActivity extends AppCompatActivity {
     if (shouldShowRationale) {
       // 3a. Rationale needed: Make rationale UI visible.
       Log.d(TAG, "startRequestPermissionFlow: Rationale needed. Making UI visible.");
+      // TODO: As per Android guidelines, in an educational UI, explain to the user why the app
+      //  requires this permission. In this UI, include a "cancel" or "no thanks" button.
+      // For now, just show the rationale message with an OK button.
       rationaleUIRoot.setVisibility(View.VISIBLE);
     } else {
       // 3b. No rationale needed: Directly request permission.


gerhardol:
If the permission is denied, the guidlines at least for phone states that you should not ask again.

The phone app is a little annoying here. I implemented that the app is not starting GPS unless all the time is ticked, to avoid error reports from users where the phone shut down GPS access after an hour or so. But that should maybe have been a snackbar or something in the GUI and maybe a default note.

roberi:
> If the permission is denied, the guidlines at least for phone states that you should not ask again.

Ok, I see. I tried to follow the [Workflow for requesting permissions](https://developer.android.com/training/permissions/requesting#workflow_for_requesting_permissions) on Android Developers, but I must have misunderstood something. Rereading them, I think the "error" is in my ActivityResultCallback. In their example, if the permission is not granted, they only "explain" to the user that the feature is unavailable because of the denial:

```java
if (isGranted) {
  // Permission is granted. Continue the action or workflow in your
  // app.
} else {
  // Explain to the user that the feature is unavailable because the
  // feature requires a permission that the user has denied. At the
  // same time, respect the user's decision. Don't link to system
  // settings in an effort to convince the user to change their
  // decision.
}
```

In my code, I show the rational UI if not granted, which has the "explanation" and an OK button that requests the permission again:

```java
if (isGranted) {
  sendPermissionResultAndFinish(true);
} else {
  if (shouldShowRequestPermissionRationale(permissionToRequest)) {
    rationaleUIRoot.setVisibility(View.VISIBLE);
  } else {
    sendPermissionResultAndFinish(false);
  }
}
```

Perhaps we should just send false instead:

```java
if (isGranted) {
  sendPermissionResultAndFinish(true);
} else {
  sendPermissionResultAndFinish(false);
}

// or simply
// sendPermissionResultAndFinish(isGranted);
```

Then, the next time the user starts the app, the rational UI will be shown to "explain" why the permission is needed. If still denied, there will be no more "explaining".

gerhardol:
> Perhaps we should just send false instead:

Seem reasonable, no deep thoughts

There are reasons permissions for GPS and battery savings are annoying (if they are sufficient I am not the one to answer...), the notifications should not be.

roberi:
I think this change better aligns the permission request workflow with the Android guidelines. 

https://github.com/user-attachments/assets/71f0ea84-597d-4899-b95b-fa7d7f027995

The "explanations" can still be improved if needed.

gerhardol:
Rebase and use this in the upcoming release?
There are some lint issues in this PR. Two are for versions, the third is below. I can fix, but not verify.


C:\dev\gc\runnerup\wear\src\main\res\layout\activity_request_permission.xml:23: Error: Possible overdraw: Root element paints background @color/black with a theme that also paints a background (inferred theme is @style/Theme_AppCompat_NoActionBar) [Overdraw]
    android:background="@color/black"
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

gerhardol:
I rebased this in my repo, have not fixed the issue
@roberi, can you take a look and update your PR?

@jonasoreland If no response from @roberi I may open a PR for this - but you will have to test

jonasoreland:
@gerhardol : I haven't been tracking this...but if you give me a branch to test I will...(on the watch I assume?)

gerhardol:
> @gerhardol : I haven't been tracking this...but if you give me a branch to test I will...(on the watch I assume?)

rebased again, see roberi_fix/wear-post-notifications in my fork

jonasoreland:
update: tested, and the permission request seems to work fine! (i though only tested "allow"...hmm)

ok to push from me!


related update: the actual wear app...has areas that can be improved
the actual handling of the notification etc is not great...
this i discovered before...and I regularly force stop+start the wear app, this fixes the problems
but is obviously not a create user experience...

gerhardol:
> update: tested, and the permission request seems to work fine! (i though only tested "allow"...hmm)

I will prepare a new PR with this contents, unless @roberi updates
The Wear app need to be updated so it is not stuck on the Play reviews too...

jonasoreland:
can we merge ? (i intend to fix some wear issues...and would like to have this merged...)

gerhardol:
> can we merge ? (i intend to fix some wear issues...and would like to have this merged...)

I will cleanup and make a PR

gerhardol:
Replaced by #1337

PR #1270, opened 2025-08-17
Modify GPX/TCX to export laps even if distance = 0

jonasoreland:
Currently only for sports that does not have GPS.

### DIFF
--- a/app/src/main/org/runnerup/export/format/ExportOptions.java
+++ b/app/src/main/org/runnerup/export/format/ExportOptions.java
@@ -16,6 +16,7 @@
  */
 
 package org.runnerup.export.format;
+import org.runnerup.workout.Sport;
 
 public class ExportOptions {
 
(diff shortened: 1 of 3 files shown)


jonasoreland:
again @roberi ?

roberi:
Looks good. The method shouldExportLap(..) could be public static.

When exporting an activity, at what point is the Sport known? If it's available before building the ExportOptions, would it be good to include Sport as part of the options (like isStrava, garminExtensions,  and accuracyExtensions)?

jonasoreland:
> Looks good. The method shouldExportLap(..) could be public static.

yea...but maybe it will be different for different exporters...not sure...

> When exporting an activity, at what point is the Sport known? If it's available before building the ExportOptions, would it be good to include Sport as part of the options (like isStrava, garminExtensions, and accuracyExtensions)?

it is not know before building ExportOptions...
it is know when reading the "ACTIVITY" table that contains the sport...

I'll merge as is...thanks!

PR #1272, opened 2025-08-19
remove package from AndroidManifest.xml

jonasoreland:
When I build I get a warning that it should be removed. I don't remember how this work...why it was there, and why it should no longer be there.

But since we build both org.runnerup and org.runnerup.debug from same manifest it "feel right"

### DIFF
--- a/app/AndroidManifest.xml
+++ b/app/AndroidManifest.xml
@@ -15,7 +15,7 @@
   ~  along with this program.  If not, see <http://www.gnu.org/licenses/>.
   -->
 
-<manifest package="org.runnerup"
+<manifest
           xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:tools="http://schemas.android.com/tools">
 


gerhardol:
This was needed as we appended ".debug" (and .free) to the original package, maybe also for intent filters.
So import workout (more?) should be tried too.

jonasoreland:
sorry, I don't understand what you are saying

"was needed", does that mean that it is no longer needed?
or do you mean that it is still needed ?

Thanks!

/Jonas

gerhardol:
> "was needed", does that mean that it is no longer needed?
or do you mean that it is still needed ?

I do not know if it is needed now but it was required before. 
But package is not used in so many situations, should be testable.

jonasoreland:
cool, thanks...
will try to investigate!

/Jonas

jonasoreland:
Update: I looked though code, there are 3 variants:
a) context.getPackageName()
b) org.runnerup.common.BuildConfig.applicationIdFull
c) org.runnerup.BuildConfig.APPLICATION_ID

In my debug build...they are all org.runnerup.debug

---

a) 
b) applicationIdFull is used as "targetPacket" in 2 xml files,
in intents names (Constants)

c) BuildConfig.APPLICATION_ID is used in ActivityProvider and WorkoutFileProvider

---

I'm honestly not sure if my change introduce any problems...but I don't think so,
the warning says that the field is ignored anyway.

---

PTAL

jonasoreland:
I'll merge this.

PR #1273, opened 2025-08-19
Misc cleanup in preparation for sport without gps

jonasoreland:


jonasoreland:
@roberi : again ?

app/src/main/org/runnerup/tracker/component/TrackerGPS.java
```diff
@@ -84,16 +86,36 @@ private Integer parseAndFixInteger(
...
     if (ContextCompat.checkSelfPermission(this.tracker, Manifest.permission.ACCESS_FINE_LOCATION)
         != PackageManager.PERMISSION_GRANTED) {
       mWithoutGps = true;
     }
     try {
-      LocationManager lm = (LocationManager) context.getSystemService(Context.LOCATION_SERVICE);
+      var lm = locationManager;
       SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
       frequency_ms = parseAndFixInteger(preferences, R.string.pref_pollInterval, "1000", context);
+      mLastLocation = getLastKnownLocation(lm);
```
roberi, line 118: Will everything work as intended if null is returned (i.e. no location provider has a last known location)? This is perhaps something that will never happen on a real device?

jonasoreland:
thanks for great review, i had only ever tested it on real phone...
my laptop is not so powerful...that the emulator almost crashes it...
so I mostly test on real phone...which is good and bad...

ptal ?

roberi:
Looks good. Now there will be at least a "RunnerUp" location as a fallback.

PR #1276, opened 2025-08-25
Fix 2 usages of deprecated sportEntries-array

jonasoreland:
Found during review of pr#1274.

This patch removes 2 usages of now deprecated sportEntries array.

### DIFF
--- a/app/src/main/org/runnerup/widget/TitleSpinner.java
+++ b/app/src/main/org/runnerup/widget/TitleSpinner.java
@@ -21,6 +21,7 @@ import android.content.Context;
 import android.util.AttributeSet;
 import android.view.LayoutInflater;
 import android.widget.AdapterView;
+import android.widget.ArrayAdapter;
 import android.widget.LinearLayout;
 import android.widget.Spinner;
 import android.widget.SpinnerAdapter;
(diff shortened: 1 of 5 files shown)


jonasoreland:
@roberi : from pr#1274

jonasoreland:
a comment is that it would be nice to do this auto-magically,
e.g. by introducing something like "runnerup:entriesReferences="@array/sportEntriesReference"
which would be a an array of reference between the DB_VALUE and the text id.

but that feels out of scope for now...

app/src/main/org/runnerup/widget/TitleSpinner.java
```diff
@@ -165,4 +166,11 @@ public void setOnSetValueListener(SpinnerInterface.OnSetValueListener listener)
   public void setOnCloseDialogListener(SpinnerInterface.OnCloseDialogListener listener) {
     mPresenter.setOnCloseDialogListener(listener);
   }
+
+  // Instead of android:entries="@array/anArray"
+  public void setArrayEntries(String[] entries) {
+    var adapter = new ArrayAdapter<CharSequence>(
+        getContext(), android.R.layout.simple_spinner_item, entries);
+    mSpinner.setAdapter(adapter);
```
roberi, line 174:
```suggestion
    adapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
    mSpinner.setAdapter(adapter);
```

The left image is without the call to `setDropDownViewResource`. 

Image

roberi (approved):
Looks good.

I prefer the look of the drop-down views when calling `setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item)` on the adapter. But ok either way.

PR #1277, opened 2025-08-29
Graph-x-axis

jonasoreland:
This patch makes it possible to use time (instead of distance) as X-axis in graphs...and makes it modifyable.
It also removes pace graph if there is no pace data.

Both of these features are planned for sport without gps, i.e Treadmill. But in theory, one can imagine useing time instead of distance as x-axis also for e.g. running. Maybe it should be a switch ? (in a future version!)

### DIFF
--- a/app/src/main/org/runnerup/util/Formatter.java
+++ b/app/src/main/org/runnerup/util/Formatter.java
@@ -185,11 +185,6 @@ public class Formatter implements OnSharedPreferenceChangeListener {
     return null;
   }
 
-  public String getElapsedUnit(Format target) {
-    return resources.getString(
-        org.runnerup.common.R.string.metrics_elapsed_unit);
-  }
-
   public static boolean getUseMetric(Resources res, SharedPreferences prefs, Editor editor) {
     boolean _km;
     String unit = prefs.getString(res.getString(R.string.pref_unit), null);
--- a/app/src/main/org/runnerup/util/GraphWrapper.java
+++ b/app/src/main/org/runnerup/util/GraphWrapper.java
@@ -58,10 +58,10 @@ public class GraphWrapper implements Constants {
     double getX(double distance, double time_ms);
   };
 
+  private final XAxis distanceXAxis;
+  private final XAxis timeXAxis;
   boolean useDistanceAsX = true;
-  XAxis distanceXAxis;
-  XAxis timeXAxis;
-  XAxis xAxis;
+  private XAxis xAxis;
 
   /** Called when the activity is first created. */
   @SuppressLint("ObsoleteSdkInt")
--- a/common/src/main/res/values/strings.xml
+++ b/common/src/main/res/values/strings.xml
@@ -23,7 +23,6 @@
   <string name="metrics_elapsed_m">m</string>
   <string name="metrics_elapsed_min">min</string>
   <string name="metrics_elapsed_s">s</string>
-  <string name="metrics_elapsed_unit">hh:mm</string>
   <string name="metrics_heartrate">HR</string>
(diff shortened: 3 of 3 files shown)


roberi:
I have not looked at the code yet, only some initial testing on an emulator (new installation).

For a short Treadmill activity in my test, about 10 seconds, the label for the x-axis in the HR graph is somewhat misleading.

image

The values are seconds, but the hh:mm label suggests that they are minutes.

Perhaps not a big deal since in practice no user will (ever?) record activities this short.

The same "issue" also exists for other activities, like a short run (the values are in meters, but the label says km):

image

Perhaps use labels like "distance" and "time". Or ignore for now.

roberi:
When adding a manual activity, regardless of sport, the graph tab (DetailActivity) is just empty (no Pace- or HR graph). It makes sense in a way, as there are no recorded values, but wouldn't a "No data" message of some kind be nice to show to the user?

roberi:
This is perhaps not something related to this specific PR, but something we missed in earlier "no GPS" related PRs.

If I record a Treadmill activity, the Lab tab (DetailActivity)  shows incorrect data:

image

Adding a manual Treadmill activity, the lap info is correct.

jonasoreland:
> I have not looked at the code yet, only some initial testing on an emulator (new installation).
> 
> For a short Treadmill activity in my test, about 10 seconds, the label for the x-axis in the HR graph is somewhat misleading.
> 
> image
> The values are seconds, but the hh:mm label suggests that they are minutes.
> 
> Perhaps not a big deal since in practice no user will (ever?) record activities this short.
> 
> The same "issue" also exists for other activities, like a short run (the values are in meters, but the label says km):
> 
> image
> Perhaps use labels like "distance" and "time". Or ignore for now.



> I have not looked at the code yet, only some initial testing on an emulator (new installation).
> 
> For a short Treadmill activity in my test, about 10 seconds, the label for the x-axis in the HR graph is somewhat misleading.
> 
> image
> The values are seconds, but the hh:mm label suggests that they are minutes.
> 
> Perhaps not a big deal since in practice no user will (ever?) record activities this short.
> 
> The same "issue" also exists for other activities, like a short run (the values are in meters, but the label says km):
> 
> image
> Perhaps use labels like "distance" and "time". Or ignore for now.

done

jonasoreland:
> When adding a manual activity, regardless of sport, the graph tab (DetailActivity) is just empty (no Pace- or HR graph). It makes sense in a way, as there are no recorded values, but wouldn't a "No data" message of some kind be nice to show to the user?

I guess...can I skip for now ??

jonasoreland:
> This is perhaps not something related to this specific PR, but something we missed in earlier "no GPS" related PRs.
> 
> If I record a Treadmill activity, the Lab tab (DetailActivity) shows incorrect data:
> 
> image
> Adding a manual Treadmill activity, the lap info is correct.

How is that incorrect ?
I manually "Lap" when running on the treadmill, the Lap info seems fine.
(I also plan to add a "auto lap based on time" to complement the existing "based on distance")

roberi:
> How is that incorrect ?
> I manually "Lap" when running on the treadmill, the Lap info seems fine.
> (I also plan to add a "auto lap based on time" to complement the existing "based on distance")

The image shows a stopped/saved activity. Instead of `0 m` and `--:-- /km`, shouldn't the lap info be `1.00 km` and `00:14 /km` (like in the app bar)?

roberi:
> > When adding a manual activity, regardless of sport, the graph tab (DetailActivity) is just empty (no Pace- or HR graph). It makes sense in a way, as there are no recorded values, but wouldn't a "No data" message of some kind be nice to show to the user?
> 
> I guess...can I skip for now ??

ok

jonasoreland:
thanks!

PR #1278, opened 2025-09-03
fix: Use correct SharedPreferences for target coaching settings

### LINKED ISSUE #1275: WorkoutBuilder uses incorrect SharedPreferences for target coaching settings
The `addAudioCuesToWorkout` method in `WorkoutBuilder` incorrectly uses an audio cue-specific SharedPreferences to read settings for "Target coaching" (preference keys `pref_target_pace_moving_average_seconds` and `pref_target_pace_grace_seconds`). These settings are actually stored in the default SharedPreferences. This results in the app always using the hardcoded default values (`20` and `30`, respectively) for these parameters instead of the user-configured values from the workout settings screen.

Steps to Reproduce:

1. Go to the app's workout settings.
2. Set a custom value for "Target feedback moving average (s)" (e.g., 10 seconds).
3. Set a custom value for "Target feedback grace period (s)" (e.g., 5 seconds).
4. Ensure "Target Coaching" audio cue (cueinfo_target_coaching) is enabled in the audio cue settings being used.
5. Start a workout that utilizes coaching.
6. After the initial 20-second grace period, audio feedback is still given every 30 seconds (instead of 5 seconds).

Update: The issue does not occur when using the default audio cue.

roberi:
The target coaching settings (keys `pref_target_pace_moving_average_seconds` and `pref_target_pace_grace_seconds`) were being ignored when a non-default audio cue was used. This was because these settings were being incorrectly read from the audio-cue-specific SharedPreferences instead of the default SharedPreferences where they are stored.

This commit updates the `addAudioCuesToWorkout` method in `WorkoutBuilder` to accept the default SharedPreferences instance in addition to the audio-cue-specific SharedPreferences. This allows the method to correctly apply the user-configured target coaching values.

Fixes #1275.

### DIFF
--- a/app/src/main/org/runnerup/view/StartFragment.java
+++ b/app/src/main/org/runnerup/view/StartFragment.java
@@ -689,7 +689,7 @@ public class StartFragment extends Fragment implements TickListener, GpsInformat
     }
     if (w != null) {
       WorkoutBuilder.prepareWorkout(getResources(), pref, w);
-      WorkoutBuilder.addAudioCuesToWorkout(getResources(), w, audioPref);
+      WorkoutBuilder.addAudioCuesToWorkout(getResources(), w, audioPref, pref);
     }
     return w;
   }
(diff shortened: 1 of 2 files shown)


jonasoreland:
LGTM

jonasoreland (approved):

Issue #1281, opened 2025-09-14
Can't unstop a pause activity


jonasoreland:
1) start activity
2) press pause
3) press stop
4) press resume

expect: back at RunActivity with workout paused
what happens: It goes to RunActivity and then directly back to DetailActivity

gerhardol:
I believe the current behavior is the correct.
I see no reason Pause and Stop behave differently, remove one of them.

jonasoreland:
making stop and pause the same is a valid discussion to have, i think
current is good, i.e. them being different

this fixes the buggy behaviour...
IF stop and pause was the same that would be moot...but it is currently NOT

gerhardol:
I rather change so they are the same

jonasoreland:
> I rather change so they are the same

So you said. And I can repeat, this is about a bug in the existing feature.

I like the distinction between pause/stop, but I guess you can:
a) make a different bug about merging them and rally support
b)  make a patch that merge them, and rally support

---

thanks/Jonas

roberi:
> So you said. And I can repeat, this is about a bug in the existing feature.

I agree, this is a bug that needs fixing.

> making stop and pause the same is a valid discussion to have

I prefer the current distinction between pausing and stopping. If I know I'm just going to make a short stop (a pause) and then continue my workout, I have no need to leave RunActivity and go to DetailActivity. On the other hand, if I know I'm done with my workout and press stop, I want to be taken to the DetailActivity and start writing some notes.

PR #1282, opened 2025-09-14
Fix resume of a paused+stopped activity

jonasoreland:
fixes https://github.com/jonasoreland/runnerup/issues/1281

### DIFF
--- a/app/src/main/org/runnerup/view/RunActivity.java
+++ b/app/src/main/org/runnerup/view/RunActivity.java
@@ -387,7 +387,8 @@ public class RunActivity extends AppCompatActivity implements TickListener {
   private final OnClickListener newLapButtonClick = v -> workout.onNewLapOrNextStep();
 
   private void updateView() {
-    if (mTracker.getState() == TrackerState.STOPPED) {
+    boolean isPaused = workout != null && workout.isPaused();
+    if (mTracker.getState() == TrackerState.STOPPED && !isPaused) {
       doStop();
     } else {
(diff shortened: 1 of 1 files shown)


jonasoreland:
@roberi ping ?

gerhardol (commented):
-0  As discussed in the issue, I prefer the current functionality but do not oppose a merge

jonasoreland:
👍

gerhardol (approved):
merge for now and we can change later, if we agree on that

roberi:
It kind of works... pressing resume (from paused + stopped state) now takes the user back to the RunActivity. But the activity isn't actually resumed (as in time starts ticking again). The user needs to press the resume button in RunActivity as well for the activity to start. E.g., two presses on a resume button in total.

Perhaps a wanted behaviour, but at least for me, I would like the activity to really be resumed. What do you think?

jonasoreland:
> It kind of works... pressing resume (from paused + stopped state) now takes the user back to the RunActivity. But the activity isn't actually resumed (as in time starts ticking again). The user needs to press the resume button in RunActivity as well for the activity to start. E.g., two presses on a resume button in total.
> 
> Perhaps a wanted behaviour, but at least for me, I would like the activity to really be resumed. What do you think?

this is how it was intended to work, yes.
i'll submit this...and we can carry on with more.
ok ?

roberi (approved):

jonasoreland:
thanks!

PR #1293, opened 2025-11-29
Add timebased autolap, so far to workouts without gps

jonasoreland:


app/src/main/org/runnerup/view/StartFragment.java
```diff
@@ -190,6 +194,7 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
     Context context = requireContext();
     mDB = DBHelper.getWritableDatabase(context);
     formatter = new Formatter(context);
+    prefs = PreferenceManager.getDefaultSharedPreferences(context);
```
gerhardol: The context is ot fixed, so I do not think this can be shared,

app/src/main/org/runnerup/workout/WorkoutBuilder.java
```diff
@@ -666,6 +666,15 @@ public static void prepareWorkout(Resources res, SharedPreferences prefs, Workou
     }
   }
 
+  public static void setAutolapTime(Workout w, double seconds) {
+    for (StepListEntry s : w.getStepList()) {
+      if (s.step.getIntensity() == Intensity.ACTIVE
```
gerhardol: Why limit to these states?

gerhardol (commented):
why only in basic and for non GPS?
enable globally instead?

Issue #345, opened 2015-10-20
Heart Rate Device Reconnection (BTLE)


jerriclynsjohn:
After the pairing process, every-time I navigate to the default activity it tries connecting to Hear Rate Monitor (BTLE), the following happens:

1. Initially the device is connected and shows the success toast message "Connected to HRM HRM1"
2. Navigate to another activity and back, the HRM is not connected, as the heart icon shows "?"
3. Manually turn off and on the Bluetooth of the android device, and then navigate to default activity again will reconnect the HRM (BTLE)
4. Same is applicable every-time you quit the application, and then resume. 

Which means, when the Bluetooth is manually turned on /off the problem is solved, this could be an android BTLE API bug (which is apparently buggy as hell) or should this be something obvious. I've not looked into the code yet, it will be highly be appreciated if someone could give me a head-start if that's necessary.

jerriclynsjohn:
When the application is open and GPS and BT can be disabled or enabled from the drawer.
1. When GPS and BT is disabled, the button shows "Enable GPS" and no HRM icon except the name
2. When GPS is enabled and BT is disabled, the button shows "Waiting for GPS" which later changes to "Start Activity" and no HRM icon except the name
3. When GPS is enabled and BT is enabled, the button shows "Start Activity" and no HRM icon except the name

Now navigate to another activity and come back again, the HRM icon is shown and gets connected.
Now again navigate to another activity and come back again, the HRM icon is shown and is not connected until you manually switch off and on the android BT then refresh to the activity.

Ideally the BTLE reconnection should work seamlessly and If I'm not mistaken by the functionality the HRM icon should be shown if both GPS and BT is enabled.

jerriclynsjohn:
@jonasoreland I think there is some default bug associated with Android API, I'll try test it out with other libraries to see if we can take care of this issue.

gerhardol:
The app now have audio cues if HR connection is lost/restored.

jonasoreland:
fixed by pr #1319

PR #1295, opened 2025-12-07
Add "feature" to debug log HRM when running

jonasoreland:
my HRM spuriously disconnect, very annoying!


jonasoreland:
This is work in progress for https://github.com/jonasoreland/runnerup/issues/345

PR #1307, opened 2026-01-14
fix: special lint config if no mapbox key

gerhardol:
Lint (and thereby GitHub Action) failed if MapBox was not enabled.
This adopts the lint configuration, if no mapbox token for the lib download, a lint config adapted for osmdroid is used.


gerhardol:
I could build the original PR in my GH repo, this failed. Then this is OK here, fails in my repo. (That is due to MAPBOX though.)
Merging this to continue.

PR #1308, opened 2026-01-15
Various fixes

gerhardol:
Follow up to #1304 with various fixes.
This is also a base for #1305 
Rebase merge

Review commit by commit, describes each change.


jonasoreland (approved):
this is sort of impossible to review...or ?
anyway, the part i looked at LGTM

gerhardol:
> this is sort of impossible to review...or ? anyway, the part i looked at LGTM

commit by commit it can be reviewed. But it is not simple. There is a pattern though for each commit.

The first and the last two could be separated, but they are not really changing anything that the user notice.

PR #1309, opened 2026-01-15
feat: mean-sea-level adjustment with Android built-in

gerhardol:
Based on #1308 

Use the (beta) androidx libraries to adjust the WGS84 elipsoid
to EGM2008 se level elevation.
The functionality is built in to Android 14 and later and available in
androidx libraries (that may require that the libs are downloaded).
The previous implementation used a custom EGM96 model bundled with the app.

This raises the min SDK from 21 (Android 5) to 23 (A6). 3 of 4149 users will not get an update. (Note that there are 4 users still on Android 4, with older versions.)
It may be required to raise minsdk further.

Note: This model may have a bigger diff to EGM2008 than the current EGM96. Some random checks indicates that they diff up to 2.5m (more likely that this is accuracy than changed models). This is still much lower than GPS elevation error. As I expect most apps to use this implementation, RU displays the same as others. (Likely just visible if you export GPX and compare point by point.)

### DIFF
--- a/app/lint-baseline.xml
+++ b/app/lint-baseline.xml
@@ -13,14 +13,14 @@
     </issue>
 
     <issue
-        id="UnusedAttribute"
-        message="Attribute `singleLineTitle` is only used in API level 26 and higher (current min is 23)"
-        errorLine1="            android:singleLineTitle=&quot;false&quot;"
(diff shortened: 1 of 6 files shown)


gerhardol:
Fixed some issues.
Removed the settings for offsetting altitude from pressure from GPS, pressure elevation does not make sense without it.
This is hard to test with mock elevation providers, I have to run outside to see the effects...

PR #1310, opened 2026-01-18
AGP 9.0.0

gerhardol:
Android Gradle Plugin update
Rebase merge

* fix: replace deprecated LocalBroadcastManager
Untested in Wear

* fix: remove Pebble support
The integration for the old Pebblekit has not been working for several
years and the pebblekit lib now has deprecated features.
As enableJetifier could not be set to false (see below)
it could stay but the demand seem low.

* fix: agp 9.0.0
Changes to support agp9, including fixing deprecations
enableJetifier could not be set to false due to com.jjoe64:graphview

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -79,7 +79,7 @@ jobs:
             **/build/outputs/logs/
 
       - name: Test
-        run: ./gradlew testBuildTypesUnitTest :app:testLatestDebugUnitTest testDebugUnitTest
+        run: ./gradlew test
 
       - name: Upload test logs
         if: always()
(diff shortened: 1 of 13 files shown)


gerhardol:
Force pushed a fix for the F-Droid build in GifHub

gerhardol:
> funnily enough, the page describing downsides with LocalBroadcastManager says that one is that developers confuse it with global broadcast :) Which this patch modifies the code to use...i.e. opposite what they want.
> 
> I'm not 100% sure that will work for Wear, but let's merge it and I'll test.

Yes, I plan to prepare a 2.10.0.0 beta in the weekend that can be tested with. There are other basic changes too.
I used Gemini to convert, did not try to find a solution myself. (Hopefully more knowledgeable than a random dude, but not always correct.)

PR #1318, opened 2026-02-07
Modify distance picket when using Miles

jonasoreland:
Before patch you enter x miles and y meters
After patch you enter x miles and y% miles (0-99)

The new format is the only natural way...
meters and miles doesn't match at all.


gerhardol (approved):
'have not run but seem ok

PR #1319, opened 2026-02-07
Redo reconnect feature for HR

jonasoreland:


gerhardol:
The first to be done for HR is to modernize the API. It currently uses the 4.3 BluetoothAdapter.startLeScan() instead of 5.0  BluetoothLeScanner.startScan() 

I asked Gemini to make the changes but it does not compile.
I suggest this is done first.

jonasoreland:
if you get a version that works, I can test it...
the version here mostly works...which is what I care about :)

jonasoreland:
fwiw, I checked into 5.0 BluetoothLeScanner.startScan
and concluded that moving to that is fine...but it will not affect the reconnect behavior...
i.e. it' s unrelated...

jonasoreland:
I've been using this personally now for a couple of month.
It fixes the reconnect problem encountered without this change.

A noticeable problem about testing this is that I don't know when the disconnect
happens...it can go days without any disconnect, only to a day later get 3 disconnects
on one run...really weird...

Anyway, I think is good enough to merge now!

PTAL

jonasoreland:
re squash: sure, np

maybe try copilot on this too ?
it did find a bug, but didn't fix it :)
maybe it find something here too ?
wdyt?

gerhardol:
@copilot review 

Requires copilot pro license to listen to instructions. Sometimes useful.

PR #1283, opened 2025-09-18
Use system file picker (SAF) for DB export/import

roberi:
Previously, database exports were saved to app-private storage, making it difficult (if not impossible) for users to access the database file outside of RunnerUp. This PR refactors the database import and export functionality from using app-private storage to using the Storage Access Framework (SAF). This framework allows users to interact with a system picker to select files from shared storage, not requiring any system permissions. For RunnerUp, users may now choose where their database backups are saved and accessed from.

Fixes #1083 and #1153.

TODOs (separate PRs):
1. Add confirmation prompt before overwriting existing DB when importing (#1089).
2. Backup the current DB before overwriting.
3. Enable direct DB file opening from file managers via AndroidManifest intent filter.

PR #1323, opened 2026-02-07
Use system file picker (SAF) for DB export/import

gerhardol:
Replaces #1283 by @roberi 
Some todo fixed, a little cleanup
TODO 2,3,4 in the original PR ignored

Translations to be updated after merge


gerhardol:
Will include this in a 2.11

PR #1335, opened 2026-03-23
WIP: Fix wear incorrectly dismiss notification

jonasoreland:
1) It seems like posting on wear does not wakeup ListenerService (as it would if same was posted on phone...weird)
2) It seems like keeping notification show works fine while the MainActivity is running, good! Much easier.

This is tested on my Pixel Watch 3.
I'll test it a couple of weeks and see if it seems ok.


wear/src/main/java/org/runnerup/view/MainActivity.java
```diff
@@ -51,6 +51,7 @@
 
 public class MainActivity extends Activity
     implements Constants, ValueModel.ChangeListener<TrackerState> {
+  private static final String TAG = "RU:MainActivity";
```
gerhardol, line 54: Why RU: for the tags?

wear/src/main/java/org/runnerup/view/MainActivity.java
```diff
@@ -68,6 +69,7 @@ public class MainActivity extends Activity
   @Override
   protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
+    Log.i(TAG, "onCreate");
```
gerhardol, line 72: These are WIP, shouldbe at least v
I removed many of them in the phone app, it was hard to follow normal flow. You may want to do that too.
Since I do not have any Wear device, I do not much debugging and leave this to you.

wear/src/main/java/org/runnerup/service/ListenerService.java
```diff
@@ -131,26 +132,28 @@ public void onDataChanged(DataEventBuffer dataEvents) {
...
+      Log.i(TAG, "ListenerService.onPeerDisconnected: " + peer.getId());
     }
   }
 
   private void maybeShowNotification() {
-    Log.d(getClass().getName(), "mainActivityRunning=" + mainActivityRunning +
-                       ", phoneApp=" + phoneApp +
-                       " ,phoneRunning=" + phoneRunning +
-                       " ,trackerState=" + trackerState);
-    if (mainActivityRunning == Boolean.TRUE) {
```
gerhardol, line 150: so this is the only change?

gerhardol (approved):

wear/src/main/java/org/runnerup/view/MainActivity.java
jonasoreland: easy to grep for...

wear/src/main/java/org/runnerup/service/ListenerService.java
jonasoreland: yes

PR #1354, opened 2026-05-19
Override Speed unit for a specific activity

lecler-i:
Hello,

I've created this little patch to be able to change the "Speed unit" for a specific activity.
This allow me to be in "pace" for my running, but be in "speed" for my biking (that make more sense for me).

This have been generated by AI but the code look clean to me, feel free to tell me right off if you don't even wanna look at it, i would understand.

I've ran it on my phone, and is working as expected.

Cf screen recording showing the new option in the settings : 

https://github.com/user-attachments/assets/8d0ef155-5939-4c13-b5c0-77c0e6c131ce


gerhardol:
Speed should be set by default for cycling related, is configuration really needed?

Issue #1167, opened 2023-10-03
Feature request: elevation gain


ngiret:
Hello,
It would be nice to indicate the elevation gain (positive and negative) after the activity is finished.
Thanks!

LorandMarton:
+1

aardvark0420:
Ive been trying to display elevation data for each km in the graph. Not sure if there is a conflict with the sensors (tried selecting either or barometer or GPS) I selected in the settings or if the app is not capable of displaying it at all? I was under the impression it was..but perhaps I am wrong? The only graph I see is the pace graph.

gerhardol:
There are graphs for pace and HR, not implemented for other types of data.

Issue #229, opened 2015-04-04
No altitude tracking function


g-allen:
I was looking for a compact, basic running app and RunnerUp looked like just what I wanted. I tried it out yesterday and it works well for me. (I'm 67 years old and run just to keep fit.)
Two suggestions however:
1. I live in a hilly area and was disappointed to find no altitude tracking. I'd like to see a graph of up and down hill (ideally superimposed on the speed graph!)
2. The map function required me to install Google Play Services, a huge 30 Mb app, which rather defeated the purpose of looking for a compact running app (my phone is short of memory). Isn't there some other way to access Google Maps? (I'm non a technical expert).
Best wishes, and thanks for your hard work.

jonasoreland:
Hi,

1) Altitude is tracked...but not (yet) graphed

1&2) The easiest to see maps and altitude (and other stuff) is to upload it
to one site
and look at it on the web...RunnerUp supports plenty of them

/Jonas

paradix:
I have been doing some changes to the graphs just for fun and wanted to create a graph similar to what you can find on Endomondo. Part of this is also the altitude,
When I find time I can show some screen shots or create a pull request.

g-allen:
Thank you, good work!
I think your app is great, keep it up!
Please let me know when these functions are incorporated in a new version.
Geoffrey

Il 07/04/2015 alle 10.49, paradix ha scritto:
>
> I have been doing some changes to the graphs just for fun and wanted 
> to create a graph similar to what you can find on Endomondo. Part of 
> this is also the altitude,
> When I find time I can show some screen shots or create a pull request.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/jonasoreland/runnerup/issues/229#issuecomment-90468674>.
>

-- 
Via S. Tommaso 13, I-25049 Pilzone (BS), Italy
tel. (+39) 030-523425, +39 348-8130057

g-allen:
Thank you.

Sorry to bother you again but... I uploaded my data to Facebook, but 
where is it now? I can't find any trace of it on FB!!!


Il 05/04/2015 alle 3.24, Jonas Oreland ha scritto:
> Hi,
>
> 1) Altitude is tracked...but not (yet) graphed
>
> 1&2) The easiest to see maps and altitude (and other stuff) is to 
> upload it
> to one site
> and look at it on the web...RunnerUp supports plenty of them
>
> /Jonas
>
[...]

PR #1357, opened 2026-05-23
Add elevation graph

lynxize:
Adds a simple elevation graph to the graph tab.

Since having 2 or 3 graphs on the same page can get a little crowded, I put the elevation graph behind a preference that defaults to false. It would probably be better to make the entire graphs tab scrollable to fit, but I'm very new to android stuff and didn't want to make too many changes.

I don't have any workouts with heart rate data so I wasn't able to test how it behaves with all 3 graphs.

Related issues: 
- https://github.com/jonasoreland/runnerup/issues/229 
- https://github.com/davidedelvento/runnerup/issues/25
- https://github.com/jonasoreland/runnerup/issues/1167 (not solved)

### DIFF
--- a/app/src/main/org/runnerup/util/GraphWrapper.java
+++ b/app/src/main/org/runnerup/util/GraphWrapper.java
@@ -45,8 +45,9 @@ import org.runnerup.view.HRZonesBar;
 import org.runnerup.workout.SpeedUnit;
 
 public class GraphWrapper implements Constants {
-  private final GraphView graphView;
-  private final GraphView graphView2;
+  private final GraphView velocityGraphView;
+  private final GraphView hrGraphView;
+  private final GraphView elevationGraphView;
 
   private final LinearLayout graphTab;
   private final HRZonesBar hrzonesBar;
@@ -120,9 +121,9 @@ public class GraphWrapper implements Constants {
 
     loadParam = new LoadParam(context, mDB, mID);
 
-    graphView = new GraphView(context);
-    graphView.setTitle(formatter.formatVelocityLabel());
-    graphView
+    velocityGraphView = new GraphView(context);
+    velocityGraphView.setTitle(formatter.formatVelocityLabel());
+    velocityGraphView
         .getGridLabelRenderer()
         .setLabelFormatter(
             new DefaultLabelFormatter() {
@@ -135,17 +136,19 @@ public class GraphWrapper implements Constants {
                 }
               }
             });
-    graphView.getGridLabelRenderer().setVerticalAxisTitle(formatter.getVelocityUnit(context));
-    graphView.getGridLabelRenderer().setHorizontalAxisTitle(xAxis.label());
+    velocityGraphView
+        .getGridLabelRenderer()
+        .setVerticalAxisTitle(formatter.getVelocityUnit(context));
+    velocityGraphView.getGridLabelRenderer().setHorizontalAxisTitle(xAxis.label());
     // enable zoom
-    graphView.getViewport().setScalable(true);
-    graphView.getViewport().setScrollable(true);
+    velocityGraphView.getViewport().setScalable(true);
+    velocityGraphView.getViewport().setScrollable(true);
 
-    graphView2 = new GraphView(context);
-    graphView2.setTitle(context.getString(org.runnerup.common.R.string.Heart_rate));
-    graphView2.getGridLabelRenderer().setVerticalAxisTitle("bpm");
-    graphView2.getGridLabelRenderer().setHorizontalAxisTitle(xAxis.label());
-    graphView2
+    hrGraphView = new GraphView(context);
+    hrGraphView.setTitle(context.getString(org.runnerup.common.R.string.Heart_rate));
+    hrGraphView.getGridLabelRenderer().setVerticalAxisTitle("bpm");
+    hrGraphView.getGridLabelRenderer().setHorizontalAxisTitle(xAxis.label());
+    hrGraphView
         .getGridLabelRenderer()
         .setLabelFormatter(
             new DefaultLabelFormatter() {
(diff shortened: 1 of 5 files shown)


gerhardol:
This could be added as a second graph instead, in both hr/pace graphs.

Note that thedependency providing the graphs need to be replaced. It is not compatible with jetifier.
The dep could maybe be fixed too.

app/res/values/pref_keys.xml
```diff
@@ -103,6 +103,7 @@
     <string name="pref_step_countdown_active">pref_step_countdown_active</string>
     <string name="pref_step_countdown_time">pref_step_countdown_time</string>
 
+    <string name="pref_show_elevation_graph">pref_show_elevation_graph</string>
```
gerhardol: Remove the settings, not really needed. The view should have scrolling though.

app/src/main/org/runnerup/util/GraphWrapper.java
```diff
@@ -45,8 +45,9 @@
 import org.runnerup.workout.SpeedUnit;
 
 public class GraphWrapper implements Constants {
-  private final GraphView graphView;
-  private final GraphView graphView2;
+  private final GraphView graphView; // pace
+  private final GraphView graphView2; // heart rate
+  private final GraphView graphView3; // elevation
```
gerhardol: Maybe rename the others too, but do not follow bad naming...
```suggestion
  private final GraphView elevationGraphView; // elevation
```

app/src/main/org/runnerup/util/GraphWrapper.java
```diff
@@ -161,6 +162,26 @@ public String formatLabel(double value, boolean isValueX) {
     graphView2.getViewport().setScalable(true);
     graphView2.getViewport().setScrollable(true);
 
+    graphView3 = new GraphView(context);
+    graphView3.setTitle(context.getString(org.runnerup.common.R.string.Elevation));
+    // graphView3.getGridLabelRenderer().setVerticalAxisTitle("alt");
```
gerhardol: remove from the values
getElevationUnit() need to be added. There is no metrics_distance_ft, so metrics_distance_m is the only option now.
```suggestion
    graphView3.getGridLabelRenderer().setVerticalAxisTitle(formatter.getElevationUnit(context));
```

app/src/main/org/runnerup/util/GraphWrapper.java
```diff
@@ -498,8 +543,7 @@ public void complete(final GraphView graphView) {
       graphView.getViewport().setMaxX(graphView.getViewport().getMaxX(true));
       graphViewData.setOnDataPointTapListener(
           (series, dataPoint) -> {
-            String msg =
-                String.format(
+            String msg = String.format(
```
gerhardol: Revert changes like this
This project uses google-java-format (not enforced in checks), this violates that format and will be changed 
I will add a GitHub action to check this

Can be run locally:
`find . -name \*.java -exec java -jar google-java-format-1.35.0-all-deps.jar --replace {} +`

app/src/main/org/runnerup/util/GraphWrapper.java
lynxize: fixed (I renamed the others)

app/src/main/org/runnerup/util/GraphWrapper.java
lynxize: `formatElevation()` seems to convert between meters and feet itself, and for me the graph is in feet.

I removed the unit from the data and added "feet"/"meters" to the axis title instead
(I'm not sure if that's preferred over "ft" and "m" but to me the abbreviations looked funny as an axis label)

app/src/main/org/runnerup/util/GraphWrapper.java
lynxize: Oh, alright, good to know! I ran it through google-java-format and it seems to have undone most of those changes. Sorry about that.

app/src/main/org/runnerup/util/GraphWrapper.java
```diff
@@ -226,18 +252,16 @@ private GraphProducer doLoadGraphInBackground(LoadParam params) {
...
-      graphTab.addView(graphView, new LayoutParams(LayoutParams.MATCH_PARENT, 0, 0.5f));
-
-      graphTab.addView(graphView2, new LayoutParams(LayoutParams.MATCH_PARENT, 0, 0.5f));
-    }
+    graphData.complete(velocityGraphView);
+    graphTab.removeView(velocityGraphView);
+    graphTab.removeView(hrGraphView);
+    graphTab.removeView(elevationGraphView);
+
+    LayoutParams layout = new LayoutParams(LayoutParams.MATCH_PARENT, 800);
```
lynxize, line 260: I'm not exactly sure what the best way to size the graphs is, now that they're inside a `ScrollView`, this was sort of a placeholder

gerhardol: Maybe this should be minimalsize for what is visible.
But it works OK on my Pixel8 and emulator SmallPhone.

gerhardol (approved):
Thanks
The graphs could be improved in many other ways, overlaying in the same graph (at least two) and have more data.
But the graph component need to be replaced soon, should be done after that.

gerhardol:
Under is needed in the graph view, the bottom is covered by the navigation pill.

lynxize:
I added some padding to the ScrollView (wasn't sure how to "correctly" handle it, but it seems to work and shouldn't interfere with the nav bar anymore) and fixed an id issue

gerhardol:
DetailedActivity.java had handling for inset: ViewCompat.setOnApplyWindowInsetsListener
so the lat change should not have been needed
But we go with this
The unit for the x axis should be removed too (was not changed in this pr)

gerhardol (approved):

gerhardol:
I pushed some changes to my branch that cleans up the inset and unit handling, as a second update to cleanup xaxis.
Tried to push to your branch, but not allowed.
The first three are fixups, the fourth is a cleanup of x axis and unit in general

Please review the three first commits and add them to your PR.
I added #1364 with x-axis etc. I could merge your changes in that PR

lynxize:
> Tried to push to your branch, but not allowed.

Oops, I don't know why I had that set but you should be able to push to my branch now.

At this point, might as well just move to #1364 though

PR #1361, opened 2026-06-06
fix: add format check to GitHub actions

gerhardol:
GitExtensions builtin to see GH builds, links

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -101,6 +101,16 @@ jobs:
             **/build/reports/lint-results-release.html
             **/build/reports/lint-results-latestRelease.html
 
+      - name: Check Java code style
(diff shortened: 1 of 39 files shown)


gerhardol:
@jonasoreland should the format test be reported as an error like this?
It could be frustrating for users (but the message has instructions).
If a warning only you have to explicitly look at the report though.

jonasoreland:
if the "./gradlew spotlessApply" just works...then an error is ok...
my $0.05

gerhardol:
> if the "./gradlew spotlessApply" just works...then an error is ok... my $0.05

It works, but you need to have Java 17 installed, which may require e.g. pointing to the Android Studio installation. So a few steps.
But then I keep it as gating, it is possible to merge anyway right now.

PR #1362, opened 2026-06-10
Add volume button controls option

ghost:
This option maps volume up to new lap and volume down to toggle pause.
It is off by default as it takes over the volume buttons.

### DIFF
--- a/app/src/main/org/runnerup/view/RunActivity.java
+++ b/app/src/main/org/runnerup/view/RunActivity.java
@@ -33,6 +33,7 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.os.IBinder;
 import android.util.Log;
+import android.view.KeyEvent;
 import android.view.LayoutInflater;
 import android.view.MotionEvent;
 import android.view.View;
@@ -209,7 +210,6 @@ public class RunActivity extends AppCompatActivity implements TickListener {
     ViewUtil.Insets(findViewById(R.id.start_view), true);
 
     showOnLockScreen(showOnLockScreen);
-
     if (keepScreenOn) {
       getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
     } else {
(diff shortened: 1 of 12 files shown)


app/src/main/org/runnerup/view/RunActivity.java
```diff
@@ -391,20 +391,52 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
   }
 
-  private final OnClickListener pauseButtonClick =
-      v -> {
-        if (workout == null) {
-          // "should not happen"
-          return;
-        }
+  @Override
+  public boolean onKeyDown(int keyCode, KeyEvent event) {
```
gerhardol, line 433: What I recall from the documentation, only media apps would receive the volume events. But it seem to wok,

app/src/main/org/runnerup/view/RunActivity.java
```diff
@@ -391,20 +391,52 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
-        if (workout.isPaused()) {
-          workout.onResume(workout);
-        } else {
-          workout.onPause(workout);
-        }
-        setPauseButtonEnabled(!workout.isPaused());
-      };
+    if (volumeButtonControls &&
+            Arrays.asList(KeyEvent.KEYCODE_VOLUME_UP, KeyEvent.KEYCODE_VOLUME_DOWN).contains(keyCode)) {
+      return true;
```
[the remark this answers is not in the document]
ghost: Done

app/src/main/org/runnerup/view/RunActivity.java
ghost: They are treated as standard Android key codes that any app can override, but they still work as volume keys when the screen is off.

ghost:
I have made it on by default to increase awareness, and since you can change the volume by turning the screen off (and ofc turn it off if it isn't useful, but I think most people will find it useful). I can undo this if needed.

gerhardol:
> I have made it on by default to increase awareness, and since you can change the volume by turning the screen off (and ofc turn it off if it isn't useful, but I think most people will find it useful). I can undo this if needed.

I think it should be off by default

gerhardol (approved):

gerhardol:
thanks

PR #1363, opened 2026-06-19
fix: show light status bar icons

ghost:
Setting the status bar appearance to dark makes the icons light, giving them adequate contrast from the dark gray background. Some similar functions are also updated.


gerhardol:
Seem fine, but how do I set light status bars?

ghost:
This doesn't change the status bar background colour, but tells Android that the status bar is dark, and therefore the icons should be white to contrast with the background.

Because of EdgeToEdge, the status bar background colour seems to be the same as the app's background colour, so for light status bars you would need a light background colour in the app and also `setAppearanceLightStatusBars(true)` for black icons.

Screen comparison

gerhardol:
I do not see the problem on my Pixel 8 w Android 17 (and before that on 16) or on A14.
But rebase on master, reformat code (with .gradlew spotlessApply) as this is changed in master.

PR #1364, opened 2026-06-19
fix: graph distance adaptive formatter

gerhardol:
First four is #1357 and some cleanups to that
Will merge these together regardless

Declutter the graph axis, consistently use km/miles in units.
Remove elevation decimals.
Cleanup some context use in Formatter
Include units in graph popups

### DIFF
--- a/app/res/layout/detail.xml
+++ b/app/res/layout/detail.xml
@@ -189,13 +189,21 @@
 		      android:layout_height="match_parent" />
 		  </LinearLayout>
 
-                  <LinearLayout
+                <ScrollView
                     android:id="@+id/tab_graph"
                     android:layout_width="match_parent"
-                    android:layout_height="wrap_content"
-                    android:orientation="vertical"
-                    android:weightSum="1"
-                    android:layout_margin="@dimen/activity_margin"/>
+                    android:layout_height="match_parent"
+                    android:scrollbars="none"
+                    android:clipToPadding="false">
+
+                    <LinearLayout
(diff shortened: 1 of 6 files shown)


lynxize:
I just tested this on my own device, and it seems to work well. The only issue I noticed was the pace graph's y-axis units displaying as `/mi` instead of `min/mi` 

image

gerhardol:
> I just tested this on my own device, and it seems to work well. The only issue I noticed was the pace graph's y-axis units displaying as `/mi` instead of `min/mi`

That was intentional, that is how pace is displayed elsewhere in the app.

Added a one ocommit for the first part, copilot comments and one for the second part with the xaxis changes etc.
Intend to merge soon, two commits.

Some notes:
* Formatter handling of Distance is even more complicated now, need to be improved. Maybe new alternatives to TXT_LONG
* miles is still substandard, even if better with this PR. Changing the X dataseries to actual data will improve, but complicate other aspects

I looked a little into replacing GraphView with something supported. 
https://github.com/philjay/mpandroidchart is most commonly used it seems but is mostly unmaintained but there are supported forks.
Another alternative is https://github.com/patrykandpatrick/vico, actively developed (and supports Jetpack, not in the plan). Less features but limited features needed.

gerhardol:
I have been working on an update with com.github.AppDevNext:AndroidChart, an update to mpandroidchart, not ready yet.
Merging this for now.

Thanks @lynxize !

PR #1376, opened 2026-08-12
fix: restore "Are you sure?" string value

roberi:
Fixes an issue introduced in 6489c52 where the key name was used instead of the intended value:

image


gerhardol (approved):

gerhardol:
lint fails, will update in a separate PR

PR #1378, opened 2026-08-13
Add initial Wear OS heart rate provider support

### LINKED ISSUE #1366: Use Smartwatch (e.g. Pixel Watch) as HR sensor
Hi, I try to use my Pixel Watch 2 (Wear OS 6 and 7) as HR sensor for RunnerUP. If I try to find the Watch in HR Monitor I cannot find it.
I CAN find my watch if I activate "Connected Fitness" on my watch and the watch also sees an incoming request from my phone and says "connected" BUT the RunnerUP app never received the HR. Did anyone get that working? I would love to provide logs and so on but I don`t know how

roberi:
This PR introduces initial support for using a Wear OS device as a heart rate provider in RunnerUp.

https://github.com/user-attachments/assets/eda9f3e0-b09c-49b2-ada6-70cb407ff760

### Phone app
- Added a new **Wear OS devices** option to `HRSettingsActivity`. This allows users to discover and connect to paired Wear OS devices running the RunnerUp Wear app.
- Added a new heart rate provider, `WearHRProvider`, which implements the existing `HRProvider` interface. By following the same pattern as the Bluetooth and ANT+ providers, the integration requires only minimal changes to the core application logic.

### Wear app

Two new services have been added to the Wear module:

#### `HeartRateService`

Responsible for collecting heart rate and battery information from the watch.

- Heart rate data is obtained through `SensorManager` using `Sensor.TYPE_HEART_RATE`.
- Battery status is monitored through a `BroadcastReceiver` listening for `Intent.ACTION_BATTERY_CHANGED`.

#### `HeartRateListenerService`

Extends `WearableListenerService` and acts as the communication layer between the phone and the watch.

- Receives START and STOP commands from the phone via the _Wearable Data API_.
- Starts and stops `HeartRateService` as needed.

## Compatibility note

The Wear module currently targets `minSdk 25`, which makes it possible to support older Wear OS devices. Because of this, the implementation uses the traditional _SensorManager API_ instead of the newer _Health Services API_, which requires Wear OS 3 (API 30) or later.

## Testing

The feature has been tested using both Android Studio emulators and physical devices (Pixel Watch 4 running Wear OS 7, paired with Pixel 9 Pro running Android 17). More than 50 workouts completed successfully. No stability issues observed, provided that the RunnerUp Wear app remains in the foreground (see limitations below). Battery consumption is approximately 20% per hour of activity on the Pixel 4.

Additional testing was performed by [Toby2988](https://github.com/Toby2988) as described in issue #1366.

## Known limitations
The watch app currently needs to remain in the foreground to ensure reliable heart rate delivery.

## Follow-up work

The following improvements are planned in subsequent PRs:

- Convert _HeartRateService_ into a foreground service (or merge it with the existing foreground notification) to ensure reliable operation when the display is off or another app is active.
- Improve error reporting and recovery during device discovery and connection establishment, including watch-side permission failures.
- Address the new READ_HEART_RATE requirements introduced on Wear OS 6 (API 36). The current implementation only suppresses the manifest warning through `tools:ignore="HealthConnectPolicy"`.


gerhardol (commented):
Brief look, no comments.
Note that Play updates of Wear have been rejected the last releases. Jonas did not get the updates approved (I do not recall the motivation right now).

roberi:
Thanks for your note. I will contact Jonas to see if there is something I can help him with.

hrdevice/build.gradle
```diff
@@ -60,8 +60,11 @@ repositories {
 }
 
 dependencies {
+    implementation project(':common')
+    
     implementation "androidx.appcompat:appcompat:${rootProject.ext.appcompat_version}"
     implementation "androidx.preference:preference:${rootProject.ext.preference_version}"
+    implementation "com.google.android.gms:play-services-wearable:${rootProject.ext.googlePlayServicesWearableVersion}"
```
gerhardol, line 67: Also need to add variant to latest and to common ittself, or at least raise an error if not latest
```suggestion
    if (rootProject.ext.enableWear) {
        latestImplementation "com.google.android.gms:play-services-wearable:${rootProject.ext.googlePlayServicesWearableVersion}"
    }
```

common/src/main/java/org/runnerup/common/util/Constants.java
```diff
@@ -274,5 +288,11 @@ interface RunInfo {
     interface TrackerState {
       String STATE = "state";
     }
+
+    interface Capability {
```
gerhardol, line 292: There must be a version, so the implementation can be changed

common/src/main/res/values/strings.xml
```diff
@@ -206,6 +206,7 @@
     <string name="Experimental_HRM_devices">Experimental HRM devices</string>
     <string name="Paired_BLE_devices">Paired BLE devices</string>
     <string name="Experimental_features">Experimental features</string>
+    <string name="Wear_OS_devices">Wear OS devices</string>
```
gerhardol, line 209: ?
Not generic
```suggestion
    <string name="Wear_OS_devices">RunnerUp Wear OS devices</string>
```

hrdevice/src/org/runnerup/hr/WearHRProvider.java
```diff
@@ -0,0 +1,368 @@
...
+        String path = messageEvent.getPath();
+        Log.d(TAG, "onMessageReceived: " + path);
+
+        if (Constants.Wear.Path.MSG_HEART_RATE.equals(path)) {
+          byte[] payload = messageEvent.getData();
+          try {
+            hrValue = Integer.parseInt(new String(payload));
+            hrTimestamp = System.currentTimeMillis();
+            hrElapsedRealtime = SystemClock.elapsedRealtimeNanos();
+            Log.d(TAG, "onMessageReceived: hrValue: " + hrValue);
```
gerhardol, line 342: This is too verbose, change before merge. Applies to a lot of printouts in this module.

wear/src/main/java/org/runnerup/hr/HeartRateListenerService.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (C) 2014 robert.jonsson75@gmail.com
```
gerhardol, line 2: some various years in other files too...
```suggestion
 * Copyright (C) 2026 robert.jonsson75@gmail.com
```

wear/src/main/java/org/runnerup/hr/HeartRateService.java
```diff
@@ -0,0 +1,355 @@
...
+ * by {@link HeartRateListenerService} in response to commands from the phone.
+ */
+public class HeartRateService extends Service implements SensorEventListener {
+  private static final String TAG = "HeartRateService";
+
+  // API 36+ requires READ_HEART_RATE. We intentionally use the literal permission
+  // string rather than HealthPermissions.READ_HEART_RATE because referencing
+  // android.health.connect classes may trigger class-loading failures on older
+  // Wear OS versions where those classes are unavailable.
+  private static final String READ_HEART_RATE_PERMISSION =
```
gerhardol, line 54: check version instead, easier to cleanup

hrdevice/build.gradle
roberi: I wasn't exactly sure what you meant, but I gave it a try. With the new commit, the build has an error. Please [have a look](https://github.com/jonasoreland/runnerup/pull/1378/changes/77d457de023d4e593ffc9223896ae8bbd2617ce9#diff-a2e69d39c47873ff474b04619dd77d0d8afa89a25383e03638a9bd4a51d35723).

gerhardol: hrdevice do not have more than one flavor, a new is not required (sorry for being unclear, I copy&pasted)
The source code related to wear must be separated, with dummy code if no wear (f-droid).
The lint error will be fixed by removing latest, the non free you will see with f-droid

gerhardol:
Without a wear device, I cannot test this (the wear emulator has not worked well when I have tried it).
The impact on the phone app is minimal, the logic is in wear. Only browsed wear, see no problem.
@jonasoreland can you take a look and try it?
Otherwise I will just approve it.

roberi:
> ... (the wear emulator has not worked well when I have tried it).

Compared to year(s) ago, I find it much easier now to pair phone and watch AVDs in Android Studio. You may already be doing this, but use a phone image with Play support, sign in to the Play Store, install the Pixel Watch companion app, and use the "pairing assistant" in Android Studio.

https://developer.android.com/training/wearables/get-started/connect-phone

gerhardol:
> > ... (the wear emulator has not worked well when I have tried it).
> 
> Compared to year(s) ago, I find it much easier now to pair phone and watch AVDs in Android Studio. You may already be doing this, but use a phone image with Play support, sign in to the Play Store, install the Pixel Watch companion app, and use the "pairing assistant" in Android Studio.

No arguments against, but remember my old experience and I let you with real devices do the testing....

jonasoreland:
1) I'm fine with just approving, and I'll test "running"

the movie is awesome!

2) I think "The watch app currently needs to remain in the foreground to ensure reliable heart rate delivery."
is a serious limitation...since that is almost never how I use the watch app...

3) I have one feature request,
I have a "real" hrm band...it's like 10x better than the watch, but I only use it when running intervals...(max 1-2 a week).
can we make a "HRM priority list" ?
so that it first check my "real" hrm, and then fallback to the watch if that is not available ?
or something like that ??

/Jonas

gerhardol:
>     3. I have one feature request,
>        I have a "real" hrm band...it's like 10x better than the watch, but I only use it when running intervals...(max 1-2 a week).
>        can we make a "HRM priority list" ?
>        so that it first check my "real" hrm, and then fallback to the watch if that is not available ?
>        or something like that ??

Most watches have kind of "any" paired selected. Garmin tries to combine the built in optical with one of(?) the belts.
A prio list is not much harder.
Separate PR.

PR #1379, opened 2026-08-16
Fix TCX export for non-GPS activities (#1337)

### LINKED ISSUE #1337: Upload of Gym Activity without GPS to Strava never works
Device: Unihertz Titan 2
Software: Android 16, RunnerUp installed from Play store, Version v2.11. 0. 1

Issue:
Hello, I use the app not just for runs, but also to record gym strength training sessions using the gym profile. This works fine and I can see the training entry afterwards including time and my heart rate. Those activities never show in Strava though. I've tried both manual upload and the auto option. I don't get any errors, the activities just never shows up on the other side. The upload works for my runs and walks without any problem.

Edit: I also tried manually exporting the tcx file and importing via the Strava website, there I get the error that the time information is missing from the file.

Best regards Johannes

roberi:
Fixes #1337

Non-GPS activities, such as Gym workouts, could be exported without lap data, resulting in incomplete TCX files that Strava rejected for missing activity time information.

In `TCX.exportLaps()`, the call to `ExportOptions.shouldExportLap()` passed lap time and distance values in the wrong order. As a result, laps from activities with a duration but no distance were incorrectly filtered out during export.

After swapping the time and distance arguments in the `shouldExportLap()` call, Gym and other non-GPS activities can now be successfully uploaded to Strava.

**Note**! There is still a separate issue related to how Strava interprets non-GPS activities when GPS-related fields are present in the exported TCX file, but this is outside the scope of this PR.

### DIFF
--- a/app/src/main/org/runnerup/export/format/TCX.java
+++ b/app/src/main/org/runnerup/export/format/TCX.java
@@ -300,7 +300,7 @@ public class TCX {
     double totalDistance = 0;
     while (lok) {
       if (exportOptions.shouldExportLap(
-          sport.getDbValue(), /* distance= */ cLap.getFloat(1), /* time= */ cLap.getLong(2))) {
+          sport.getDbValue(), /* distance= */ cLap.getFloat(2), /* time= */ cLap.getLong(1))) {
         long lap = cLap.getLong(0);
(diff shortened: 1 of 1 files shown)


gerhardol (approved):
have not run

PR #1383, opened 2026-09-04
Fix bundled workout copy failure on fresh installations

roberi:
This PR fixes an issue that prevented bundled workout files from being copied during a fresh installation.

The initial suspicion came after reading several recently opened issues from @ByteBreeze233. Screenshots from the *Manage Workouts* activity showed that the four bundled workouts were missing (left picture), even though they should have been copied automatically during app installation (right picture).

image

In `MainLayout.handleBundled()`, the destination directory (`dstDir`) was incorrectly initialized using `dstBase`, which points to the parent directory, instead of `dst`, which represents the intended target subdirectory.

As a result, required subdirectories such as `app_workouts` were never created before attempting to copy the bundled workout files. When the application tried to copy the workout files into the non-existent directory, it threw a `FileNotFoundException`.

I've tested the fix on my Google Pixel 6 running Android 16 and on an Android 10 emulator in Android Studio.

### DIFF
--- a/app/src/main/org/runnerup/view/MainLayout.java
+++ b/app/src/main/org/runnerup/view/MainLayout.java
@@ -315,7 +315,7 @@ public class MainLayout extends AppCompatActivity {
 
         if (!isFile) {
           // The request is hierarchical, source is still on a directory level
-          File dstDir = new File(dstBase);
+          File dstDir = new File(dst);
           //noinspection ResultOfMethodCallIgnored
           dstDir.mkdir();
           if (!dstDir.isDirectory()) {


app/src/main/org/runnerup/view/MainLayout.java
```diff
@@ -315,7 +315,7 @@ private void handleBundled(AssetManager mgr, String srcBase, String dstBase) {
 
         if (!isFile) {
           // The request is hierarchical, source is still on a directory level
-          File dstDir = new File(dstBase);
+          File dstDir = new File(dst);
```
gerhardol, line 318: Blame indicates that this has been broken since 2016. Is it really so?

gerhardol (approved):
have not run

roberi:
Yes, it seems so. =(

app/src/main/org/runnerup/view/MainLayout.java
roberi, line 363: I considered adding another change in this PR.

Currently, each bundled asset is only copied once. A true value is stored in SharedPreferences for a resource, regardless of whether the copy operation completed successfully. On line 351 in MainLayout.java, the code records that the import/copy of the current asset succeeded.

However, this happens before the actual copy operation. If an exception occurs inside the try block on lines 356-358 and the asset is never copied, no further attempts will be made because the asset has already been marked as imported.

Should we consider moving line 351 (`pref.edit().putBoolean(key, true).apply()`) to after line 358 (`handleHooks(add)`), so that the resource is only marked as imported once the copy operation has completed successfully?

gerhardol: yes

PR #1385, opened 2026-09-12
fix: analyzeReleaseR8Config warnings

gerhardol:
- remove some depreciation warnings No update for wear, requires some GMS updates first

- fix some gradle 10 syntax in builds.gradle and gradle.properties

- use ivy repo for ant+ lib

This is related to optimization requirements policy, see
https://android-developers.googleblog.com/2026/08/app-quality-memory-optimization-secure-onboarding.html
The changes are really just nitpicks, there seem to be no actual changes.

PR #1387, opened 2026-09-13
fix: ble hr deprecation fixes

gerhardol:
Update code with api-13 to remove deprecated calls. (suppress for <api-33).

Support HR transmitter that do not support battery or firmware. This occurs for instance for Garmin watches set to send HR over bluetooth.

Some changes moved from #1385


hrdevice/src/org/runnerup/hr/AndroidBLEHRProvider.java
```diff
@@ -231,9 +248,9 @@ public void onConnectionStateChange(BluetoothGatt gatt, int status, int newState
         }
```
utafrali, line 248: **Missing deprecated `onDescriptorRead` override breaks Android < 13.**

The PR correctly provides both the deprecated 3-param and new 4-param overrides for `onCharacteristicChanged` and `onCharacteristicRead`. But `onDescriptorRead` only overrides the API-33 `(gatt, descriptor, status, value)` form. On Android < 13, the system calls the deprecated `(gatt, descriptor, status)` form, which now falls through to the base-class no-op — meaning `enableNotification` is never called and HR notification setup silently fails.

Fix:
```java
@Override
@SuppressWarnings("deprecation")
public void onDescriptorRead(
    BluetoothGatt gatt, BluetoothGattDescriptor descriptor, int status) {
  onDescriptorRead(gatt, descriptor, status, descriptor.getValue());
}

@Override
public void onDescriptorRead(
    BluetoothGatt gatt, BluetoothGattDescriptor btDesc, int status, byte[] value) {
  BluetoothGattCharacteristic mHRMcharac = btDesc.getCharacteristic();
  if (!enableNotification(gatt, true, mHRMcharac)) {
    reportConnectFailed("Failed to enable notification in onDescriptorRead");
  }
}
```

hrdevice/src/org/runnerup/hr/AndroidBLEHRProvider.java
```diff
@@ -78,28 +80,37 @@ public class AndroidBLEHRProvider extends BtHRBase implements HRProvider {
...
+              log("onCharacteristicChanged(" + btChar + ") != HEART_RATE ??");
               return;
             }
 
-            int length = arg0.getValue().length;
+            int length = value.length;
             if (length == 0) {
               log("onCharacteristicChanged length = 0");
               return;
             }
```
utafrali, line 107: **Potential `ArrayIndexOutOfBoundsException` in UINT16 heart rate reading.**

The bounds check only guards against `length == 0`, but:
- The UINT8 path (`value[1]`) requires `length >= 2`
- The UINT16 path (`value[2]`) requires `length >= 3`

A malformed BLE packet with `length == 1` and the UINT16 flag set will throw at `value[2]`. The old `getIntValue()` call handled this internally; manual extraction must replicate that guard.

Fix — replace the `length == 0` check with:
```java
if (length < 2) {
  log("onCharacteristicChanged: insufficient data, length=" + length);
  return;
}
```
And before the UINT16 branch add:
```java
if (isHeartRateInUINT16(value[0]) && length < 3) {
  log("onCharacteristicChanged: UINT16 flag set but length=" + length);
  return;
}
```

utafrali (commented):
The constant renames, graceful handling of missing DIS service (Garmin watch support), and battery null-check are solid improvements. However, the `onDescriptorRead` callback is missing its deprecated 3-param override, which will silently break HR notification setup on Android < 13 — the exact devices this PR claims to keep compatible.

hrdevice/src/org/runnerup/hr/AndroidBLEHRProvider.java
```diff
@@ -231,9 +248,9 @@ public void onConnectionStateChange(BluetoothGatt gatt, int status, int newState
         }
```
utafrali, line 249: **Misleading parameter name: `btChar` for a `BluetoothGattDescriptor`.**

Every other use of `btChar` in this file refers to a `BluetoothGattCharacteristic`. Using it here for a `BluetoothGattDescriptor` will confuse readers. Rename to `btDesc` or `descriptor` to match the type.

hrdevice/src/org/runnerup/hr/AndroidBLEHRProvider.java
```diff
@@ -468,18 +493,41 @@ private boolean enableNotification(
...
 
+    boolean result;
     if (onoff) {
-      clientConfig.setValue(BluetoothGattDescriptor.ENABLE_NOTIFICATION_VALUE);
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+        result =
+            gatt.writeDescriptor(clientConfig, BluetoothGattDescriptor.ENABLE_NOTIFICATION_VALUE)
+                == BluetoothStatusCodes.SUCCESS;
+      } else {
+        @SuppressWarnings("deprecation")
```
utafrali: **Duplicated API-version branching in `enableNotification` can be simplified.**

The on/off branches are structurally identical — only the notification value differs. Extract the API check to a single branch:
```java
byte[] notifValue = onoff
    ? BluetoothGattDescriptor.ENABLE_NOTIFICATION_VALUE
    : BluetoothGattDescriptor.DISABLE_NOTIFICATION_VALUE;
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
    result = gatt.writeDescriptor(clientConfig, notifValue)
        == BluetoothStatusCodes.SUCCESS;
} else {
    @SuppressWarnings("deprecation")
    boolean setValueResult = clientConfig.setValue(notifValue);
    @SuppressWarnings("deprecation")
    boolean writeDescResult = gatt.writeDescriptor(clientConfig);
    result = setValueResult && writeDescResult;
}
```
This cuts ~15 lines and removes the duplication.

hrdevice/src/org/runnerup/hr/AndroidBLEHRProvider.java
gerhardol: a variant of this

hrdevice/src/org/runnerup/hr/AndroidBLEHRProvider.java
gerhardol: changed, as per prev suggestion

PR #1390, opened 2026-09-17
refactor: migrate HistoryFragment to Activity Result API

Threeguana:
Replaces the deprecated startActivityForResult / onActivityResult pattern
in HistoryFragment with registerForActivityResult using the
StartActivityForResult contract
- FAB (add manual activity)
- List item click (open DetailActivity)
- Both callbacks restart the loader on return, matching the original behaviour
The deprecated onActivityResult override is removed entirely.
No behaviour change


roberi (approved):
Thank you for fixing the two TODOs, and for your first contribution to RunnerUp!

gerhardol:
@roberi Did you try it too?
I have not looked at how to try the change

roberi:
> @roberi Did you try it too? I have not looked at how to try the change

Yes, tried it in Android Studio with an emulator. Working like before.

PR #1391, opened 2026-09-17
fix: use Log.e instead of printStackTrace in AboutPreference

Threeguana:
Replaces two bare printStackTrace calls with Log.e


gerhardol (approved):

gerhardol:
thanks!

